PR #5, opened 2015-10-30
Move to subdirectory format

jeffmendoza:
@remi @bshaffer 

What do you think about the directory name for the current example: "storage"? It is currently being used here: https://cloud.google.com/storage/docs/authentication


jeffmendoza:
bump

jeffmendoza:
Updated storage example, no lib/storage. Keeping tests in spec/ as that is where "rake spec" looks

.travis.yml
```diff
@@ -5,12 +19,17 @@ rvm:
...
+    - vendor/bundle
 
 env:
-- PATH=$PATH:$HOME/gcloud/google-cloud-sdk/bin GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/client_secrets.json #Other environment variables on same line
+- TEST_DIR=storage
+  PATH=$PATH:$HOME/gcloud/google-cloud-sdk/bin
+  GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/client_secrets.json
+- TEST_DIR=pubsub
+  PATH=$PATH:$HOME/gcloud/google-cloud-sdk/bin
+  GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/client_secrets.json
```
bshaffer: This is interesting to use the build matrix for each sample's test suite. Wouldn't it be simpler to run them sequentially using one build?

jeffmendoza: I saw this idea and thought it was cool =)

bshaffer: While I can't say I'm a huge fan (it'll complicate the build UI by adding lots of rows), I'm not hugely against it. However, you should use the `global` keyword for env:

```yaml
env:
  global:
    - PATH=$PATH:$HOME/gcloud/google-cloud-sdk/bin
    - GOOGLE_APPLICATION_CREDENTIALS=$TRAVIS_BUILD_DIR/client_secrets.json
  matrix:
    - TEST_DIR=storage
    - TEST_DIR=pubsub
```

jeffmendoza: Ah, nice I'll do that.

Let's try the matrix, I think I prefer it to some sort of script that loops
through the dirs.

bshaffer: SGTM!

pubsub/spec/test_test.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "spec_helper"
+
+RSpec.describe "Pub/Sub test test" do
```
remi: Can we remove the `pubsub` subdirectory from this pull request?

bshaffer:
@jeffmendoza also it looks like `list_buckets.rb` appears twice. I am assuming the one in [`lib/storage`](https://github.com/GoogleCloudPlatform/ruby-docs-samples/tree/jlm/subdirs/lib/storage) should be removed?

remi:
`lib/storage` will be removed after the cloud.google.com code snippets have been updated to point to the new file locations :smile:

bshaffer:
bump :)

jeffmendoza:
Sorry, I've been at kubecon. I'll take a look tomorrow.

pubsub/spec/test_test.rb
jeffmendoza: I just wanted to make sure that the travis setup actually worked with more than one directory. I'll submit the PR with the pub sub sample I have right away.

jeffmendoza:
Updated, can we merge?

bshaffer:
@jeffmendoza I believe @remi wanted to remove the pubsub directory since no samples actually exist in there. Other than that, LGTM

PR #9, opened 2015-12-01
updates for storage example

bshaffer:
- adds 'all' region for docs
 - makes example classes consistent with bigquery classes


bshaffer:
@remi @jeffmendoza PTAL

jeffmendoza:
LGTM

PR #11, opened 2016-03-21
adds appengine hello world e2e tests

bshaffer:
@remi @jeffmendoza PTAL


bshaffer:
We should move e2e tests out of travis and onto cloud jenkens, but for simplicity we can get them running here first.

jeffmendoza:
Awesome! LGTM

PR #14, opened 2016-03-31
only build master branch in travis

bshaffer:


jeffmendoza:
Will this still build PRs from other branches in the gcp repo?

remi:
I'd like to configure Travis to build all branches of GoogleCloudPlatform/ruby-docs-samples

(No pull requests / remote repositories)

bshaffer:
@jeffmendoza yes, it will build PRs, but it won't build the branch. The reasoning is we don't want to have two tests running for each PR.

@remi wouldn't it make more sense to build for PRs but *not* on branches? That way we can still run unit tests for PRs from extra-google forks.

bshaffer:
We would need to upgrade the test suite to skip tests requiring credentials, which we don't currently have, for that to actually be possible.

remi:
Hmm.  On the one hand, it would be really nice to verify that tests for my branch pass on Travis •before* sending a Pull Request to see them fail.  On the other hand, building all branches means that we'll see failures every time one of us pushes a branch with failing code.  I'm OK with that though.

Whichever we way go, the most important thing for me is to confirm with 100% certainty that Travis never runs tests sent from external forks.  Only branches of GoogleCloudPlatform/ruby-docs-samples should ever be run.

jeffmendoza:
Travis already takes care of not supplying secrets to test runs from PR's from other repos. This is why those will all fail.

If we had any meaningful unit test (not integration) we could setup rspec to only run those if the secrets were empty

tmatsuo:
What Jeff is proposing is ideal. Also we can provide a guide for setting up
their travis build with their own projects for heavy contributors.

broady:
That's what I do for the Go tests.

They only run when certain environment variables are set. To simplify things, I set up a `SYSTEM_TESTS` "secret" env variable:
```
before_install:
- if [ $SYSTEM_TESTS ]; then
    openssl aes-256-cbc -K $encrypted_e44ee91b46d4_key -iv $encrypted_e44ee91b46d4_iv -in key.json.enc -out key.json -d;
    export GOLANG_SAMPLES_PROJECT_ID=golang-samples-tests;
    export GOOGLE_APPLICATION_CREDENTIALS="$(pwd)/key.json";
  else
    echo "Not running system tests.";
  fi
```

(the tests actually check for `GOLANG_SAMPLES_PROJECT_ID`, but I didn't want that to be a secret variable)

tmatsuo:
FYI, also maybe you can use TRAVIS_SECURE_ENV_VARS

   - TRAVIS_SECURE_ENV_VARS “true” or “false” depending on the availability
   of environment variables


PHP team simplified the contribution process with base64 encoded json file
so that they don't need to install travis tool.
https://github.com/GoogleCloudPlatform/getting-started-php/blob/master/TRAVIS.md

bshaffer:
closing in favor of #27

PR #17, opened 2016-04-20
Cloud Storage sample for App Engine flexible environment

remi:


appengine/storage/app.yaml
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START app_yaml]
+runtime: ruby
+vm: true
+entrypoint: bundle exec ruby app.rb -p 8080
+
+env_variables:
+  GCLOUD_PROJECT: <your-project-id>
```
jeffmendoza: Did you try to deploy without GCLOUD_PROJECT set? I normally have to have it set to run locally, but not here for running in app engine

jeffmendoza:
LGTM, I imagine you want to squash.

appengine/storage/app.yaml
remi: Ah, good call!  Removing

PR #28, opened 2016-04-29
Add Ruby Endpoints sample

broady:


remi:
The repo I pushed is not in sync with GCP/ruby-docs-samples.  Additionally, Googlebot is upset.  Sending a follow-up pull.

PR #35, opened 2016-08-04
Add IAM samples to pubsub.

jeffmendoza:


jeffmendoza:
@remi  PTAL

jeffmendoza:
@bshaffer I think remi is ooo, can you take a look?

pubsub/sample.rb
```diff
@@ -106,3 +106,89 @@ def list_subscriptions
   end
 end
 # [END list_subscriptions]
+
+# [START print_topic_policy]
+def print_topic_policy
```
remi: Rename method and doc region tag to `get_topic_policy`

pubsub/sample.rb
```diff
@@ -106,3 +106,89 @@ def list_subscriptions
...
+
+  policy = topic.policy
+
+  puts "Topic policy:"
+  puts policy.roles
+end
+# [END print_topic_policy]
+
+# [START print_subscription_policy]
+def print_subscription_policy
```
remi: Rename method and doc region tag to `get_subscription_policy`

remi:
2 naming nits for consistency.  Otherwise LGTM

pubsub/sample.rb
```diff
@@ -106,3 +106,89 @@ def list_subscriptions
   end
 end
 # [END list_subscriptions]
+
+# [START print_topic_policy]
```
remi: Please put doc region tags inside methods so the code sample shows the relevant code (not wrapped in a method)

jeffmendoza: All the existing tags are around the methods https://cloud.google.com/pubsub/publisher

Do you want me to change all of them?

remi: Ah, you're right!  I forgot these were live 😄 

Yeah, let's go ahead and change them.  It looks really nice without the method definition.  Eg. the way Python samples show up on that page

jeffmendoza: Sounds good, will do.

jeffmendoza:
@remi Comments addressed, ptal

PR #36, opened 2016-08-15
Update Cloud SQL sample to use Cloud SQL second generation

remi:
- Use socket instead of host
 - Update README


jeffmendoza:
LGTM

PR #38, opened 2016-08-16
[WIP] - adds skeleton, some basic methods for BigQuery

bshaffer:
This is how I plan to set up the new BigQuery samples and tests. I'll be finishing the stubbed ones tomorrow.


bigquery/query.rb
```diff
@@ -31,22 +31,28 @@ def unique_words project_id
...
         # [START run_query]
-        sql = "SELECT TOP(corpus, 10) as title, COUNT(*) as unique_words " +
-              "FROM [publicdata:samples.shakespeare]"
         results = bigquery.query sql
         # [END run_query]
 
         # [START print_results]
-        results.each do |row|
-          puts "#{row['title']}: #{row['unique_words']}"
+        results.each_with_index do |row, index|
```
remi: The index adds unnecessary complexity to listing results.  Any reason why the index is needed?  I'd love to simply see `results.each do |row|`

bigquery/query.rb
```diff
@@ -31,22 +31,28 @@ def unique_words project_id
...
+          puts "--- Row #{index+1} ---"
+          for column, value in row
```
remi: Prefer `each` over `for` (this will fail rubocop)

bigquery/query.rb
```diff
@@ -31,22 +31,28 @@ def unique_words project_id
...
+            puts "#{column}: #{value}"
```
remi: What are column and value?

In the previous sample, column data was printed out using explicit names.  That really helps me understand what is being printed as a user.  The query probably SELECTs a few columns and I'd love to see those columns printed out explicitly, eg. `row["title"]`

bigquery/spec/datasets_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require_relative "spec_helper"
+require_relative "../datasets"
+
+RSpec.describe "List datasets sample" do
+  before do
+    @sample = Samples::BigQuery::Datasets.new
```
remi: Lately we've been defining samples using simple top-level methods.  *(this isn't documented as a preferred convention... that's something I will do!)*

See:
 - Datastore [`sample.rb`](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/datastore/sample.rb)
 - Pub/Sub [`sample.rb`](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/pubsub/sample.rb)
 - Logging [`sample.rb`](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/logging/sample.rb)

These are wonderfully simple and no classes/modules are required 😄

bigquery/query.rb
bshaffer: The sample runs any query, so it just iterates and prints out the field name/value.

bigquery/spec/datasets_spec.rb
bshaffer: Ahh I see. This is nice, but there's no runnable sample with this method?

remi: Correct, we are not wrapping samples with programs

remi: (FWIW App Engine applications are implemented as applications so they can be deployed for E2E tests)

bshaffer: We could definitely do it this way but then add a main method at the bottom, a la [python samples](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/bigquery/cloud-client)

bigquery/spec/import_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    # delete the temporary dataset
+    @dataset.delete force: true
+  end
+
+  [
+    [File.expand_path("spec/data/test_data.json"), "JSON"],
+    [File.expand_path("spec/data/test_data.csv"), "CSV"],
+    ["gs://#{BUCKET_NAME}/test_data.json", "Cloud Storage JSON"],
+    ["gs://#{BUCKET_NAME}/test_data.csv", "Cloud Storage CSV"]
+  ].each do |source, type|
```
bshaffer: Is this an OK way to do data providers for rspec?

bigquery/query.rb
bshaffer: Naw, just to say what row we're on. I can remove it though.

bigquery/spec/import_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+require_relative "spec_helper"
+require_relative "../import"
+
+RSpec.describe "Import table data" do
+  before do
+    @sample = Samples::BigQuery::Import.new
+
+    # create a temporary dataset / table
+    require "gcloud"
```
remi: `require` statements should be at the top of the file and not inside methods/modules/etc.  We only put `require` statements inside of methods for snippet purposes.

bigquery/spec/import_spec.rb
remi: You can loop to generate the examples but it's not ideal.  Usually we would use Shared Examples for this:
http://rspec.info/documentation/3.5/rspec-core/#Shared_Examples_and_Contexts

As a general rule, though, I try to avoid using shared examples because it usually adds unnecessary complication

Because there isn't lots of redundant code, I wouldn't DRY this too much ... I'd do something explicit like:

```ruby
require_relative "spec_helper"
require_relative "../import"
require "gcloud"

RSpec.describe "Import table data" do

  before do
    @project_id  = ENV["GCLOUD_PROJECT"]
    @bucket_name = ENV["BUCKET"]
    @gcloud      = Gcloud.new ENV["GCLOUD_PROJECT"]
    @bigquery    = @gcloud.bigquery
    @dataset     = @bigquery.create_dataset "test_dataset_#{Time.now.to_f}"
    @dataset.create_table "test_table" do |schema|
      schema.string "name", mode: :required
      schema.string "title", mode: :required
    end
  end

  after do
    @dataset.delete force: true
  end

  it "import JSON" do
    source = File.expand_path "data/test_data.json", __dir__

    expect { import @project_id, @dataset.dataset_id, "test_table", source }.
      to output(/Data imported successfully/).to_stdout

    results = @bigquery.query "SELECT * FROM #{@dataset.dataset_id}.test_table"

    expect(results.size).to eq 3
    expect(result[0]).to include("name" => "Alice")
    expect(result[1]).to include("name" => "Bob")
  end

  it "import CSV" do
    source = File.expand_path "data/test_data.csv", __dir__

    expect { import @project_id, @dataset.dataset_id, "test_table", source }.
      to output(/Data imported successfully/).to_stdout

    results = @bigquery.query "SELECT * FROM #{@dataset.dataset_id}.test_table"

    expect(results.size).to eq 3
    expect(result[0]).to include("name" => "Alice")
    expect(result[1]).to include("name" => "Bob")
  end

  it "import JSON from Cloud Storage" do
    source = "gs://#{@bucket_name}/test_data.json"

    expect { import @project_id, @dataset.dataset_id, "test_table", source }.
      to output(/Data imported successfully/).to_stdout

    results = @bigquery.query "SELECT * FROM #{@dataset.dataset_id}.test_table"

    expect(results.size).to eq 3
    expect(result[0]).to include("name" => "Alice")
    expect(result[1]).to include("name" => "Bob")
  end

  it "import CSV from Cloud Storage" do
    source = "gs://#{@bucket_name}/test_data.csv"

    expect { import @project_id, @dataset.dataset_id, "test_table", source }.
      to output(/Data imported successfully/).to_stdout

    results = @bigquery.query "SELECT * FROM #{@dataset.dataset_id}.test_table"

    expect(results.size).to eq 3
    expect(result[0]).to include("name" => "Alice")
    expect(result[1]).to include("name" => "Bob")
  end
end
```

bigquery/spec/data/test_data.csv
```diff
@@ -0,0 +1,3 @@
+"Brent Shaffer","Master of the Deprecated"
```
remi: nit: I'd prefer to see generic names used and not see our names in the code anywhere.  eg. Alice and Bob

bshaffer:
I changed the format for `browse_table`, `query`, and `query_as_job`. Will this work?

bigquery/query.rb
```diff
@@ -0,0 +1,50 @@
...
+require "gcloud"
+
+def query project_id, sql
+  # [START build_service]
+  gcloud = Gcloud.new project_id
+  bigquery = gcloud.bigquery
+  # [END build_service]
+
+  # [START run_query]
+  results = bigquery.query sql
```
tswast: Could you show how to use standard SQL syntax and/or file bugs against gcloud-ruby if the useLegacySql parameter isn't available yet?

bigquery/query_as_job.rb
```diff
@@ -0,0 +1,49 @@
...
+require "gcloud"
+# A short sample demonstrating making a BigQuery request as a job
+# This uses Application Default Credentials to authenticate.
+# @see https://cloud.google.com/bigquery/bigquery-api-quickstart
+def query_as_job project_id, sql
+  gcloud = Gcloud.new project_id
+  bigquery = gcloud.bigquery
+
+  # [START run_query]
+  job = bigquery.query_job sql
```
tswast: Ditto for query_as_job. Could you show how to use standard SQL syntax and/or file bugs against gcloud-ruby if the useLegacySql parameter isn't available yet?

remi:
Before we add this new sample, let's update it to use the google-cloud-bigquery gem (*as part of gem renaming*)

remi:
FYI I'm working on some updates to this that I'll send for review later today

bigquery/browse_table.rb
```diff
@@ -0,0 +1,54 @@
...
+  # [START browse_table]
+  gcloud = Gcloud.new project_id
+  bigquery = gcloud.bigquery
+
+  dataset = bigquery.dataset dataset_id
+  table = dataset.table table_id
+  row_num = 0
+  page_token = nil
+  loop do
+    data = table.data :token => page_token, :max => max
```
remi: FYI

```ruby
data = table.data token: page_token, max: max
```

Don't use hash rockets `=>` unless the key of the hash must be a String or type besides a Symbol ... in those cases, use of a hash literal might not be recommended

bigquery/browse_table.rb
```diff
@@ -0,0 +1,54 @@
...
+    data.each do |row|
+      puts "--- Row #{row_num+=1} ---"
+      for column, value in row
```
remi: FYI

Use `#each` not `for`

remi:
Reformatted this code a bit to fit into the template we've been using for samples more recently.  See #53 for follow-up to this.

PR #49, opened 2016-09-01
Wrap Datastore insert and update samples in transactions

remi:
These samples actually run an upsert because they use `datastore.save`
rather than explicitly calling #insert and #update.  `datastore.save` is
the correct, simple idiomatic sample but these samples execute multiple
queries against Datastore rather than one so they should be wrapped in
transactions.  If either query fails, the entire transaction will be
rolled back and an exception will be raised.

### DIFF
--- a/datastore/sample.rb
+++ b/datastore/sample.rb
@@ -141,17 +141,20 @@ def upsert
 end
 
 def insert
+  task = nil
   gcloud = Gcloud.new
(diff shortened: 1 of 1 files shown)


remi:
Passed OK: https://travis-ci.org/GoogleCloudPlatform/ruby-docs-samples/jobs/156939430

Issue #51, opened 2016-09-16
Add guide for writing Ruby code sample applications and snippets


frankyn:
**Story:** I'm a Rubyist that wants to contribute. 

**Issue:** 
1. Rubyist clicks on **CONTRIBUTING.md** at the bottom of the page https://github.com/GoogleCloudPlatform/ruby-docs-samples. 
2. Rubyist arrives at https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/CONTRIBUTING.md.
3. Rubyist clicks on **Google Cloud Platform Samples Style Guide** and arrives at https://github.com/GoogleCloudPlatform/Template/wiki/style.html. 
4. Rubyist unknowingly clicks on  **ruby-docs-samples** and goes back to https://github.com/GoogleCloudPlatform/ruby-docs-samples. 

**Possible Solution:** Insert the Ruby Style Guide

remi:
As a contributor, I want to understand ...

 - How to add a new sample
 - How to modify an existing sample, eg. to provide a bug fix

PR #53, opened 2016-09-19
Ruby Samples for Google Cloud BigQuery (updated to use google-cloud-bigquery)

remi:
Follow-up from #38 -- this reformats #38 into for template that we've been using for recent samples

Note: project listing is not included in this pull request.  See [google-cloud-ruby #911](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/911)


bigquery/spec/bigquery_sample_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+# limitations under the License.
+
+require_relative "../datasets"
+require_relative "../tables"
+require "rspec"
+require "google/cloud"
+require "csv"
+
+RSpec.describe "Google Cloud BigQuery samples" do
+
```
bshaffer: is it standard to have a newline here?

bigquery/spec/bigquery_sample_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+      }.to output(
+        "Deleted dataset: test_dataset\n"
+      ).to_stdout
+
+      expect(@bigquery.dataset "test_dataset").to be nil
+    end
+  end
+
+  describe "Managing Tables" do
+
```
bshaffer: And here. Seems inconsistent.

bigquery/datasets.rb
```diff
@@ -0,0 +1,84 @@
...
+  # [END delete_dataset]
+end
+
+if __FILE__ == $PROGRAM_NAME
+  project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+  command    = ARGV.shift
+
+  case command
+  when "create"
+    create_dataset project_id: project_id, dataset_id: ARGV.shift
```
bshaffer: Is there any validation for a missing argument? i.e. what happens if `dataset_id` is not supplied?

bigquery/spec/bigquery_sample_spec.rb
remi: There's no hard and fast rule for it.

I usually put a space after my `describe` block, otherwise the example description is squished and doesn't stand out.  You'll see both if you browse code samples.

#fyi I specifically set a RuboCop rule to allow for this [[1](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/.rubocop.yml#L75)]

My personal rule has always been: *always add a newline before an example*

I don't always add the newline before `before` blocks, though, so the samples are inconsistent 😕 

Putting together a style guide is a high priority, see also: #51

bigquery/datasets.rb
remi: It'll blow up 😸 

I don't have any validations for any of the samples at the moment for simplicity.

I'd like to revisit in the near future*, but keep them consistently naively simple at the moment

* eg. follow-up on the more robust CLI parsing and sample template that I worked on a few weeks ago -- want to revisit soon

bshaffer:
LGTM

remi:
@bshaffer Whitespace is near and dear to the hearts of Rubyists ... so it's a super duper relevant comment 😄   I'll work on making the samples more consistent -- this was actually a really good catch

PR #58, opened 2016-09-22
Fix and simplify Language entry detection spec example

remi:


remi:
Tests [PASS](https://travis-ci.org/GoogleCloudPlatform/ruby-docs-samples/jobs/162013386)

PR #66, opened 2016-09-26
Add simple BASH script to run local integration tests

remi:


frankyn:
LGTM. Verified that script defined environmental variables were consistent with samples.

PR #86, opened 2016-10-11
Adds notes for fixing missing "project" field

gguuss:


vision/README.md
```diff
@@ -33,3 +33,12 @@ Example:
     ruby vision_samples.rb labels   /path/to/cat.jpg
     ruby vision_samples.rb landmark /path/to/grand-canyon.jpg
     ruby vision_samples.rb faces    /path/to/faces.jpg output-image.jpg
+
+Troubleshooting:
+
+If you encounter the message, "project is missing", set the configured
```
remi: Let's just add an instruction to tell users to explicitly export `GOOGLE_CLOUD_PROJECT`

Rather than troubleshooting after the fact, let's direct users to set the variable before running the sample.

gguuss: Done

PR #90, opened 2016-10-13
Refactor Logging Quickstart spec and fix by adding order clause to entry fetch

remi:
Refactor Logging Quickstart spec and fix by adding order clause to entry fetch

 - `logging.entries filter:` was not working without additional `order:` option
   - *is this intended behavior or a bug?*
 - simplify log name to `"quickstart_log_#{Time.now.to_i}"`
 - always try to delete log after spec runs, allowing for NotFound exceptions
 - refactor entry lookup into a helper method


frankyn:
quickstart passed. sample spec failed.

PR #94, opened 2016-10-24
Update AppEngine CloudSQL second generation sample

remi:
- Update doc region tags for coming doc updates
 - Refactor query onto its own line with comment for clarity in code sample snippet
 - Add spec that can be run locally to verify working sample


appengine/cloudsql/spec/end_to_end_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+
+require File.expand_path("../../../../spec/e2e", __FILE__)
+require "rspec"
+require "capybara/rspec"
+require "capybara/poltergeist"
+
+Capybara.current_driver = :poltergeist
+
+RSpec.describe "Cloud SQL on Google App Engine", type: :feature do
```
frankyn: Do we need to set RSpec. here? Looking for consistency.

frankyn:
Additional review of README:
-- Add allowed characters string, because the use of underscores in the placeholder can be misleading.
---- Ex from Console: "Use lowercase letters, numbers, or hyphens. Start with a letter".
-- I understand that **Running Locally** answers how to run it locally, but "bundle install" should be a step in the instruction set if not you can't run create_tables.rb.
-- You have two "in" on instruction #8

appengine/cloudsql/spec/end_to_end_spec.rb
remi: Yes, `RSpec.describe` is used everywhere.

frankyn: https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/1c4bbba3a542c5dd51b59be26ddfdbc7e6495faf/appengine/cloudsql/spec/cloudsql_spec.rb#L19

remi:
Done ~ ready for review @frankyn

appengine/cloudsql/spec/end_to_end_spec.rb
remi: Ah, looks like some use the global, monkey-patched method.  Kk, updated all to prefer `describe`

PR #99, opened 2016-10-27
Using timestamps in static names for Bigquery tests

frankyn:
There's an issue where Travis and Circle fail each others tests, because they're using the same static name to test with. This PR will add time to generate unique names.

@remi

### DIFF
--- a/bigquery/spec/bigquery_sample_spec.rb
+++ b/bigquery/spec/bigquery_sample_spec.rb
@@ -28,10 +28,10 @@ describe "Google Cloud BigQuery samples" do
     @bucket     = @storage.bucket ENV["GOOGLE_CLOUD_STORAGE_BUCKET"]
     @tempfiles  = []
 
-    file_time     = Time.now.to_i
-    @file_name    = "bigquery-test_#{file_time}"
-    @dataset_name = "test_dataset_#{file_time}"
-    @table_name   = "test_table_#{file_time}"
+    current_time  = Time.now.to_i
+    @file_name    = "bigquery-test_#{current_time}"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,7 @@
 ## Ruby docs samples
 
 [![Build Status](https://travis-ci.org/GoogleCloudPlatform/ruby-docs-samples.svg?branch=master)](https://travis-ci.org/GoogleCloudPlatform/ruby-docs-samples)
+[![CircleCI](https://circleci.com/gh/GoogleCloudPlatform/ruby-docs-samples/tree/master.svg?style=svg)](https://circleci.com/gh/GoogleCloudPlatform/ruby-docs-samples/tree/master)
```
remi: How come this is part of this PR?

bigquery/spec/bigquery_sample_spec.rb
```diff
@@ -28,35 +28,40 @@
     @bucket     = @storage.bucket ENV["GOOGLE_CLOUD_STORAGE_BUCKET"]
     @tempfiles  = []
 
-    # Examples assume that newly created test_dataset and test_table exist
-    delete_test_dataset!
+    file_time     = Time.now.to_i
```
remi: What does "file time" mean?

What about calling this `timestamp` or `current_time`?

frankyn: Agreed, this can be named better!

README.md
frankyn: Oops?? I merged in Master to make sure they were on the same commit. I'll fix that.

PR #102, opened 2016-10-27
Remove Travis Status Badge from README

frankyn:
@remi


remi:
If we're going to stop using Travis, can we please remove the .travis.yml as well?  I find it confusing otherwise

frankyn:
I have deleted .travis.yml

PR #107, opened 2016-11-03
Fix endpoints tests.

jeffmendoza:
Need to see if circle passes.


jeffmendoza:
Endpoints tests fixed, but other failures still present.

frankyn:
The other bugs keep breaking. Still flaky. Endpoints test looks good to me.

PR #113, opened 2016-11-16
Translate API no longer requires an API key.

jmdobry:
See 

- GoogleCloudPlatform/php-docs-samples#248
- GoogleCloudPlatform/python-docs-samples#659


jmdobry:
I don't know why the tests are failing. Help?

puneith:
Lots of stuff removed from Gemfile.lock? And I see some of the test errors like faraday missing due to that?

jmdobry:
I have no idea what I'm doing. 🐶

frankyn:
I attempted to reproduce and I was able to get passed it by including "google-cloud" gem in the Gemfile. Many gems are not installed when only using "google-cloud-translate". I'm not sure why.

frankyn:
It was a bug, and it has been fixed. I just committed an updated Gemfile.lock.

puneith:
@frankyn Build still red.

frankyn:
@puneith, the red is for flaky tests. Translate tests passed :)

jmdobry:
Thanks!

PR #115, opened 2016-11-16
Swap ports for Endpoints docker image.

jeffmendoza:
Need to swap ports to line up with samples in other languages.

### DIFF
--- a/endpoints/getting-started/Dockerfile.custom
+++ b/endpoints/getting-started/Dockerfile.custom
@@ -14,4 +14,4 @@ RUN gem install bundler && \
 
 COPY * /app/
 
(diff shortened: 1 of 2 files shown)


jeffmendoza:
note, updated `gcr.io/google-samples/echo-ruby:1.0` is pushed that listens on 8080.

remi:
Makes sense!  LGTM 👍 

For additional context, where is `Dockerfile.custom` used?  Can I find it referenced by one of the [Endpoints Quickstarts](https://cloud.google.com/endpoints/docs/quickstarts)?

jeffmendoza:
Not used explicitly. The image will be a part of the GKE quickstarts, and I wan to have the source of the Docker image in the repo.

remi:
Oh, gotcha!!  Sgtm 😄

PR #123, opened 2016-11-22
Updated region tag for list_bucket_contents_with_prefix

frankyn:
**Context**
The _list_bucket_contents_with_prefix_ sample method currently uses the region tag for _list_bucket_contents_. I'm updating the region tag to reflect the sample. 

This duplication comes up on the following page: https://cloud.google.com/storage/docs/managing-buckets#storage-list-objects-ruby


frankyn:
@jeffmendoza Thank you!

PR #124, opened 2016-11-23
Update app.yaml

piaxc:
We're trying to move towards using the connection name everywhere. It's available using gcloud now, and I've also added a step for folks to record it earlier in the procedure. CL coming soon.

### DIFF
--- a/appengine/cloudsql/app.yaml
+++ b/appengine/cloudsql/app.yaml
@@ -24,7 +24,7 @@ env_variables:
   # The instance connection name can be determined from this command:
   # `gcloud sql instances describe [INSTANCE_NAME]` or by viewing the
   # Instance details page in the Google Cloud Platform Console.
-  MYSQL_SOCKET_PATH: [YOUR_INSTANCE_CONNECTION_NAME]
+  MYSQL_SOCKET_PATH: /cloudsql/[YOUR_INSTANCE_CONNECTION_NAME]
 # [END env_variables]
 
 # [START cloudsql_settings]
--- a/appengine/cloudsql/app.yaml
+++ b/appengine/cloudsql/app.yaml
@@ -21,9 +21,8 @@ env_variables:
   MYSQL_USER: [YOUR_USER]
   MYSQL_PASSWORD: [YOUR_PASSWORD]
   MYSQL_DATABASE: [YOUR_DATABASE]
-  # The instance connection name can be determined from this command:
-  # `gcloud sql instances describe [INSTANCE_NAME]` or by viewing the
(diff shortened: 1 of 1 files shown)


appengine/cloudsql/app.yaml
```diff
@@ -21,14 +21,16 @@ env_variables:
   MYSQL_USER: [YOUR_USER]
   MYSQL_PASSWORD: [YOUR_PASSWORD]
   MYSQL_DATABASE: [YOUR_DATABASE]
-  # This path was printed to the console when you started the proxy, and is of
-  # the format: `/[DIR]/[YOUR_PROJECT_ID]:[YOUR_REGION]:[YOUR_INSTANCE_NAME]`
-  MYSQL_SOCKET_PATH: [YOUR_SOCKET_PATH]
+  # The instance connection name can be determined from this command:
+  # `gcloud sql instances describe [INSTANCE_NAME]` or by viewing the
+  # Instance details page in the Google Cloud Platform Console.
+  MYSQL_SOCKET_PATH: [YOUR_INSTANCE_CONNECTION_NAME]
```
bshaffer: the socket path needs to be `/cloudsql/[YOUR_INSTANCE_CONNECTION_NAME]`

appengine/cloudsql/app.yaml
```diff
@@ -21,14 +21,16 @@ env_variables:
   MYSQL_USER: [YOUR_USER]
   MYSQL_PASSWORD: [YOUR_PASSWORD]
   MYSQL_DATABASE: [YOUR_DATABASE]
-  # This path was printed to the console when you started the proxy, and is of
-  # the format: `/[DIR]/[YOUR_PROJECT_ID]:[YOUR_REGION]:[YOUR_INSTANCE_NAME]`
-  MYSQL_SOCKET_PATH: [YOUR_SOCKET_PATH]
+  # The instance connection name can be determined from this command:
+  # `gcloud sql instances describe [INSTANCE_NAME]` or by viewing the
+  # Instance details page in the Google Cloud Platform Console.
```
piaxc: We're making an effort to get away from the concatenation approach. How about:

    # This path was printed to the console when you started the proxy, and is of
    # the format: `/[DIR]/[YOUR_INSTANCE_CONNECTION_NAME]`

followed by what I've got in there now about how to get the connection name. Would that be ok?

frankyn: I think the concatenation approach helps understand what to look for. It could be confused with project id at first glance.

remi:
LGTM.  @bshaffer or @frankyn PTAL

frankyn:
Thanks @piaxc! LGTM.

PR #131, opened 2016-12-02
Remove old sample that was using google-api-client

remi:
Can merge after sample is no longer shown here: https://cloud.google.com/docs/authentication#code_samples


frankyn:
LGTM

PR #132, opened 2016-12-02
Syntax updates to TaskList for embedded snippets

remi:
- Simplify snippet for "the call to create the authorized Cloud Datastore
   service object" to just a require statement and variable creation
 - Use double quotes consistently
 - Remove 2 unnecessary newlines to shorten sample showing listing tasks
 - Rename new_task to add_task for consistency with other samples and to make more clear

### DIFF
--- a/datastore/tasks.rb
+++ b/datastore/tasks.rb
@@ -12,13 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# [START build_service]
-require "google/cloud/datastore"
-
 def create_client project_id
+  # [START build_service]
+  require "google/cloud/datastore"
+
   @datastore = Google::Cloud::Datastore.new project: project_id
+  # [END build_service]
 end
-# [END build_service]
 
 # [START add_entity]
 def new_task description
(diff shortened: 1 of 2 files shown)


remi:
Updated to 'Do Not Merge' ~ making additional changes to prepare for docs

datastore/tasks.rb
```diff
@@ -48,14 +49,12 @@ def mark_done task_id
 
 # [START retrieve_entities]
 def list_tasks
-  query = @datastore.query("Task").
-          order("created")
-
+  query = @datastore.query("Task").order("created")
```
jeffmendoza: Frank had the same suggestion on the original PR. We left it as is, my original comment:

> For the datastore queries, it seems like a pattern to put every .(something) on a new line
> https://cloud.google.com/datastore/docs/concepts/queries#datastore-basic-query-ruby
> there are more examples with 1 or 2 where, order, etc. on new lines

I'm ok either way, just wanted to bring it to your attention.

jeffmendoza:
LGTM

datastore/tasks.rb
remi: Thanks!  Makes sense.

Usually we do that when we go over 80 LOC.

Once you go over 80, sometimes it can be nicer to put one on each line because it looks nicer than having 3 on one line with 1 trailing on a newline, for example.

PR #134, opened 2016-12-07
adds e2e tests to run-all.sh

bshaffer:
`E2E=true` is required for the e2e tests to run

### DIFF
--- a/appengine/analytics/spec/google_analytics_spec.rb
+++ b/appengine/analytics/spec/google_analytics_spec.rb
@@ -21,6 +21,8 @@ Capybara.current_driver = :poltergeist
 
 describe "Google Analytics on Google App Engine", type: :feature do
   before :all do
+    skip "End-to-end tests skipped" unless E2E.run?
+
     @url = E2E.url
   end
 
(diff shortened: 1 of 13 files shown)


frankyn:
Thanks Brent! It's looking great :) 

I did a brief review of Circle CI and found [CircleCI-Run-357]( https://circleci.com/gh/GoogleCloudPlatform/ruby-docs-samples/357) didn't propagate an error in Datastore using Ruby 2.2.5. 

After digging around a little bit I found this: [Stackoverflow Solution](http://stackoverflow.com/questions/4667509/problem-accessing-a-global-variable-from-within-a-while-loop). The issue exists from the piped path list into the while loop found in **run-all.sh**. A workaround would either be to save the list of directories in a variable without piping or use **process substitution**. I don't think process substitution is very readable though.

bshaffer:
Good catch, I've added `set -ex` to the bash script so it fails immediately

bshaffer:
Tests now fail as expected :D

bshaffer:
Tests pass for real this time, PTAL!

frankyn:
@bshaffer, Thanks! I reviewed it and it LGTM. 

I have a question, is the following still true? I'm not sure what this is referring to in your PR. 

> Good catch, I've added set -ex to the bash script so it fails immediately

bshaffer:
@frankyn no, a later comment removed it 👍

PR #138, opened 2016-12-16
Cloud Storage encryption samples

frankyn:
This PR contains a sample for upload and downloading files with user managed encryption keys. 

We are currently blocked on a key_rotation sample because it's [not yet supported ](https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/730)in `google-cloud-ruby`.

### DIFF
--- a/storage/files.rb
+++ b/storage/files.rb
@@ -47,16 +47,14 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
 end
 
 def generate_encryption_key_base64
-  # This is a sample key generator
-  # [START generate_encryption_key_base64]
   require "base64"
+  require "openssl"
 
-  encryption_key = OpenSSL::Cipher.new("aes-256-cfb").encrypt.random_key
+  encryption_key  = OpenSSL::Cipher.new("aes-256-cfb").encrypt.random_key
   encoded_enc_key = Base64.encode64 encryption_key
-  puts "Sample encryption key: #{encoded_enc_key}"
-  # [END generate_encryption_key_base64]
-end
 
+  puts "Sample encryption key: #{encoded_enc_key}"
+end
 
 def upload_file project_id:, bucket_name:, local_file_path:,
                                            storage_file_path:
@@ -78,13 +76,13 @@ def upload_file project_id:, bucket_name:, local_file_path:,
 end
 
 def upload_encrypted_file project_id:, bucket_name:, local_file_path:,
-                          storage_file_path: nil, base64_encryption_key:
+                          storage_file_path: nil, encryption_key:
   # [START upload_encryption_file]
-  # project_id            = "Your Google Cloud project ID"
-  # bucket_name           = "Your Google Cloud Storage bucket name"
-  # local_file_path       = "Path to local file to upload"
-  # storage_file_path     = "Path to store the file in Google Cloud Storage"
-  # base64_encryption_key = "Base64 encoded AES-256 encryption key"
+  # project_id        = "Your Google Cloud project ID"
+  # bucket_name       = "Your Google Cloud Storage bucket name"
+  # local_file_path   = "Path to local file to upload"
+  # storage_file_path = "Path to store the file in Google Cloud Storage"
+  # encryption_key    = "AES-256 encryption key"
 
   require "google/cloud/storage"
 
@@ -92,20 +90,19 @@ def upload_encrypted_file project_id:, bucket_name:, local_file_path:,
 
   bucket = storage.bucket bucket_name
 
-  encryption_key = Base64.decode64 base64_encryption_key
   file = bucket.create_file local_file_path, storage_file_path,
                             encryption_key: encryption_key
 
   puts "Uploaded #{file.name} with encryption key"
 end
 
-
 def download_file project_id:, bucket_name:, file_name:, local_path:
   # [START download_file]
-  # project_id  = "Your Google Cloud project ID"
-  # bucket_name = "Your Google Cloud Storage bucket name"
-  # file_name   = "Name of file in Google Cloud Storage to download locally"
-  # local_path  = "Path to local file to save"
+  # project_id     = "Your Google Cloud project ID"
+  # bucket_name    = "Your Google Cloud Storage bucket name"
+  # file_name      = "Name of file in Google Cloud Storage to download locally"
+  # local_path     = "Path to local file to save"
+  # encryption_key = "Base64 encoded AES-256 encryption key"
 
   require "google/cloud/storage"
 
@@ -120,7 +117,7 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
 end
 
 def download_encrypted_file project_id:, bucket_name:, storage_file_path:,
-                            local_file_path:, base64_encryption_key:
+                            local_file_path:, encryption_key:
   # [START download_file]
   # project_id  = "Your Google Cloud project ID"
   # bucket_name = "Your Google Cloud Storage bucket name"
@@ -132,8 +129,6 @@ def download_encrypted_file project_id:, bucket_name:, storage_file_path:,
   storage = Google::Cloud::Storage.new project: project_id
 
   bucket = storage.bucket bucket_name
-
-  encryption_key = Base64.decode64 base64_encryption_key
 
   file = bucket.file storage_file_path, encryption_key: encryption_key
   file.download local_file_path, encryption_key: encryption_key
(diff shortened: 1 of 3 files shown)


frankyn:
@remi would you please review this. Thank you!

storage/files.rb
```diff
@@ -46,6 +46,18 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
   # [END list_bucket_contents_with_prefix]
 end
 
+def generate_encryption_key_base64
+  # This is a sample key generator
```
remi: nit: Unnecessary comment

storage/spec/files_spec.rb
```diff
@@ -20,11 +20,22 @@
...
-    @project_id      = @storage.project
-    @bucket          = @storage.bucket @bucket_name
-    @local_file_path = File.expand_path "resources/file.txt", __dir__
+    @bucket_name          = ENV["GOOGLE_CLOUD_STORAGE_BUCKET"]
+    @storage              = Google::Cloud::Storage.new
+    @project_id           = @storage.project
+    @bucket               = @storage.bucket @bucket_name
+    @local_file_path      = File.expand_path "resources/file.txt", __dir__
+    @local_enc_file_path  = File.expand_path "resources/enc_file.txt", __dir__
+    @encryption_key        = generate_encryption_key
```
remi: nit: Variable assignment alignment

storage/files.rb
```diff
@@ -46,6 +46,18 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
...
+  # [START generate_encryption_key_base64]
+  require "base64"
+
+  encryption_key = OpenSSL::Cipher.new("aes-256-cfb").encrypt.random_key
+  encoded_enc_key = Base64.encode64 encryption_key
```
remi: nit: Variable assignment alignment

storage/files.rb
```diff
@@ -46,6 +46,18 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
...
+  puts "Sample encryption key: #{encoded_enc_key}"
```
remi: nit: Add space before `puts`

storage/files.rb
```diff
@@ -65,6 +77,29 @@ def upload_file project_id:, bucket_name:, local_file_path:,
...
+  # storage_file_path     = "Path to store the file in Google Cloud Storage"
+  # base64_encryption_key = "Base64 encoded AES-256 encryption key"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+
+  bucket = storage.bucket bucket_name
+
+  encryption_key = Base64.decode64 base64_encryption_key
```
remi: Why do we ask the user for a *"Base64 encoded AES-256 encryption key"* if we have to decode it before using it?

Shouldn't we ask for a key that has not been encoded if that's what the client library expects?

storage/files.rb
```diff
@@ -84,6 +119,28 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
   # [END download_file]
 end
 
+def download_encrypted_file project_id:, bucket_name:, storage_file_path:,
+                            local_file_path:, base64_encryption_key:
+  # [START download_file]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Your Google Cloud Storage bucket name"
+  # file_name   = "Name of file in Google Cloud Storage to download locally"
+  # local_path  = "Path to local file to save"
```
remi: Missing placeholder variable for `base64_encryption_key`

storage/spec/resources/enc_file.txt
```diff
@@ -0,0 +1 @@
+Content of test enc_file.txt
```
remi: To keep things simpler, how about using the existing file.txt?

As far as I can tell, there is no reason for the encryption tests to upload/download a different file than the non-encrypted samples.

storage/files.rb
```diff
@@ -245,13 +315,18 @@ def run_sample arguments
...
-  make_public <bucket> <file>         Make a file in a bucket public
-  rename      <bucket> <file> <new>   Rename a file in a bucket
+  list         <bucket>                List all files in the bucket
+  upload       <bucket> <file>         Upload local file to a bucket
+  enc_upload   <bucket> <file> <base64_enc_key> Upload local file as an
+                                         encrypted file to a bucket
+  download     <bucket> <file> <path>  Download a file from a bucket
+  enc_download <bucket> <file> <path> <base64_enc_key> Download an encrypted
+                                         file from a bucket
+  gen_key                              Genereate a sample encryption key
```
remi: s/Genereate/Generate/

storage/files.rb
```diff
@@ -46,6 +46,18 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
   # [END list_bucket_contents_with_prefix]
 end
 
+def generate_encryption_key_base64
+  # This is a sample key generator
+  # [START generate_encryption_key_base64]
```
remi: Add test for this sample snippet.

`generate_encryption_key_base64` is untested AFAICT

storage/spec/files_spec.rb
```diff
@@ -108,6 +121,26 @@ def capture &block
...
+    expect {
+      upload_encrypted_file(project_id:            @project_id,
+                            bucket_name:           @bucket_name,
+                            local_file_path:       @local_enc_file_path,
+                            storage_file_path:     "enc_file.txt",
+                            base64_encryption_key: @encoded_enc_key)
+    }.to output(
+      "Uploaded enc_file.txt with encryption key\n"
+    ).to_stdout
+
```
remi: What happens if you try to download the file without providing an encryption key?

I noticed that the upload & download tests are exactly the same as the ones that don't use customer-supplied encryption keys.  This is great but it make me wonder: *Is there a way that we could verify the file was really encrypted?*

Not required, but I'm curious what happens.  If you download a file *without* passing the key, is an error thrown, or is the file successfully downloaded (but the content is encrypted)?  If, for example, an error is thrown, it might make the test more clear if:

 1. Upload w/ encryption key
 1. Try download, verify error thrown
 1. Download, verify content OK

Just a thought!

storage/files.rb
frankyn: I'm taking input from the CLI which wouldn't be kind to non-ASCII characters. So I ask for Base64 encoded representation. Is that too much?

storage/files.rb
frankyn: Acknowledged.

storage/files.rb
frankyn: Acknowledged

storage/files.rb
remi: Handle user input/output in the "main() method" where we call the method, not inside the snippet, itself.

For example, if we want to have a code sample that has an integer as a variable, I might...

```ruby
def add_numbers a:, b:
  # [START add_numbers]
  # a = 1  # Number to add
  # b = 2  # Number to add

  result = a + b

  puts "Result: #{result}"
  # [END add_numbers]
end

if __FILE__ == $PROGRAM_NAME
  command = ARGV.shift

  case command
  when "add"
    first_number  = ARGV.shift.to_i   # <---- convert user input here  😎
    second_number = ARGV.shift.to_i

    add_numbers a: first_number, b: second_number
  else
    puts "Unknown command: #{command}"
  end
end
```

----

**P.S.** Another way to think about it... In a CLI application, think of the main method as a "controller."  It handles user input/output.  Then it operates on models.  In our case, these are our snippets.

**Goal:** prioritize the experience of *reading* the snippets on cloud.google.com so they make the most sense.  The CLI application is secondary.

storage/files.rb
remi: Or, if this won't be part of the website, let's remove the doc region tag 😄

storage/files.rb
```diff
@@ -46,6 +46,16 @@ def list_bucket_contents_with_prefix project_id:, bucket_name:, prefix:
   # [END list_bucket_contents_with_prefix]
 end
 
+def generate_encryption_key_base64
+  require "base64"
```
remi: If this isn't part of the documentation, move the require statements to the top of the file 😄

storage/files.rb
```diff
@@ -65,12 +75,34 @@ def upload_file project_id:, bucket_name:, local_file_path:,
...
   # [START download_file]
-  # project_id  = "Your Google Cloud project ID"
-  # bucket_name = "Your Google Cloud Storage bucket name"
-  # file_name   = "Name of file in Google Cloud Storage to download locally"
-  # local_path  = "Path to local file to save"
+  # project_id     = "Your Google Cloud project ID"
+  # bucket_name    = "Your Google Cloud Storage bucket name"
+  # file_name      = "Name of file in Google Cloud Storage to download locally"
+  # local_path     = "Path to local file to save"
+  # encryption_key = "Base64 encoded AES-256 encryption key"
```
remi: `#download_file` doesn't use encryption key

storage/files.rb
```diff
@@ -260,6 +330,6 @@ def run_sample arguments
   end
 end
 
-if __FILE__ == $0
+if __FILE__ == $PROGRAM_NAME
```
remi: Nice tweak ~ conventions #ftw! 🎉

storage/files.rb
frankyn: 😅 fixing now. Thank you.

storage/files.rb
frankyn: 😅 fixing now. Thank you.

storage/files.rb
```diff
@@ -65,6 +75,27 @@ def upload_file project_id:, bucket_name:, local_file_path:,
...
+
+  storage = Google::Cloud::Storage.new project: project_id
+
+  bucket = storage.bucket bucket_name
+
+  file = bucket.create_file local_file_path, storage_file_path,
+                            encryption_key: encryption_key
+
+  puts "Uploaded #{file.name} with encryption key"
+end
```
remi: Missing `# [END upload_encryption_file]`

storage/files.rb
```diff
@@ -84,6 +115,27 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
   # [END download_file]
 end
 
+def download_encrypted_file project_id:, bucket_name:, storage_file_path:,
+                            local_file_path:, encryption_key:
+  # [START download_file]
```
remi: Should be `# [START download_encrypted_file]`

storage/files.rb
```diff
@@ -84,6 +115,27 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
...
+
+  storage = Google::Cloud::Storage.new project: project_id
+
+  bucket = storage.bucket bucket_name
+
+  file = bucket.file storage_file_path, encryption_key: encryption_key
+  file.download local_file_path, encryption_key: encryption_key
+
+  puts "Downloaded encrypted #{file.name}"
+end
```
remi: Missing `# [END download_encrypted_file]``

PR #141, opened 2017-01-23
Update generate encryption key snippet

frankyn:
Updated method `generate_encryption_key_base64` to be a code sample. CloudSite will be using this section and the `require` statements should be within the method definition. 

Summary of Changes
* Changed Base64.encode64 to Base64.strict_encode64 because Base64.encode64 adds a newline character at the end of the encoded data.  
* Moved `require` statements within the method `generate_encryption_key_base64`
* Added a unit test that checks the method generates a base64 encoded encryption key.
* Added region tags around the sample.

### DIFF
--- a/storage/files.rb
+++ b/storage/files.rb
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require "base64"
-require "openssl"
-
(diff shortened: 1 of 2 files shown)


frankyn:
I mocked the OpenSSL::Cipher and I tested for a base64 string instead. I'm open to suggestions :) 

@remi PTAL.

PR #146, opened 2017-02-13
KMS Sample Update PR

frankyn:
This pull request updates the KMS samples with the following:

* [Bug] Incorrect comments used for certain samples
* [Bug] Updated test methods with 'test_blah' -> 'blah_test' 
* [Cleanup] Updated printing full path to a specific crypto key version to printing the version number only. (File: kms.rb Line: 177)
* [Updated] README to reflect added features
* [Feature] kms_remove_member_from_cryptokey_policy
* [Feature] kms_add_member_to_keyring_policy
* [Feature] kms_enable_cryptokey_version
* [Feature] kms_restore_cryptokey_version
* [Feature] kms_set_cryptokey_primary_version

### DIFF
--- a/kms/kms.rb
+++ b/kms/kms.rb
@@ -50,7 +50,7 @@ $create_cryptokey = -> (project_id:, key_ring_id:, crypto_key:, location:) do
   # project_id  = "Your Google Cloud project ID"
   # key_ring_id = "The ID of the new key ring"
   # crypto_key  = "Name of the crypto key"
-  # location    = "The location of the new key ring"
+  # location    = "The location of the key ring"
 
   require "google/apis/cloudkms_v1beta1"
 
@@ -79,9 +79,9 @@ end
 $encrypt = -> (project_id:, key_ring_id:, crypto_key:, location:, input_file:, output_file:) do
   # [START kms_encrypt]
   # project_id  = "Your Google Cloud project ID"
-  # key_ring_id = "The ID of the new key ring"
+  # key_ring_id = "The ID of the key ring"
   # crypto_key  = "Name of the crypto key"
-  # location    = "The location of the new key ring"
+  # location    = "The location of the key ring"
   # input_file  = "File to encrypt"
   # output_file = "File name to use for encrypted input file"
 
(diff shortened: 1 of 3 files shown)


kms/kms.rb
```diff
@@ -174,18 +174,90 @@
       Cloudkms::CryptoKey.new(purpose: "ENCRYPT_DECRYPT")
   )
 
-  puts "Created version #{crypto_key_version.name} for key " +
-       "#{crypto_key} in key ring #{key_ring_id}"
+  version = crypto_key_version.name.split("/").last
```
remi: 😭

remi: Would you mind explaining the change? 😄

frankyn: crypto_key_version.name contains a the full resource path to the crypto key version. I'm just splitting the path by "/" and getting the last element, the version number.

remi: 1. What do the other samples do?
2. Is there another way to get just the version number?
3. How important is it for us to print just the number?  Will users understand if crypto_key_version.name is the full path to the version?  Is that problematic?

I liked it how it was 😛

kms/spec/kms_spec.rb
```diff
@@ -163,6 +214,31 @@ def add_test_member_to_keyring_policy project_id:, key_ring_id:, member:, role:,
...
+      key_ring_id: key_ring_id,
+      location: location
+    )
+
+    resource = "projects/#{project_id}/locations/#{location}/" +
+               "keyRings/#{key_ring_id}"
+
+    if policy.bindings
+      policy.bindings.delete_if do |binding|
+        binding.role.include? role and binding.members.include? member
```
remi: As a general rule, don't use the `and` or `or` keywords.  They are not aliases for `&&` and `||` and they can often cause unexpected hair pulling.

They are banned in the GitHub styleguide.  Search for "keywords are banned" : [GitHub Ruby Styleguide](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md)

See also:
 - http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/
 - https://github.com/bbatsov/ruby-style-guide/issues/194

Use `&&` and `||` instead.  I know... they're not as pretty... darn you Ruby!

remi:
👍  save for the open comments 😄

kms/kms.rb
frankyn: PHP - does the same thing (https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/kms/api/src/functions.php#L215)
Nodejs - does not. (https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1225)

I could just revert these changes :)

remi: I vote revert if that's cool with you

While `name.split("/").last` is super duper common and some people will get what you're doing right away...

 1. It's complexity that I don't think is very useful for the sample
 2. It's confusing to me.  When I see `name.split("/").last` I think *"huh... um.  what?  so, I get that this is splitting on slashes and getting the last bit but... why?  What's the rest of the string?"*

This is one of those times that I value *simplicity* of sample over strict *correctness*

😄

frankyn: Got it! I'll revert! :dango:

frankyn:
@remi Reverted changed PTAL!

PR #155, opened 2017-04-03
Fix issue where E2E specs cause failures even when they are disabled

remi:
See fixed example CircleCI output: https://circleci.com/gh/GoogleCloudPlatform/ruby-docs-samples/561

Failed, but no E2E failures.  1 quote exceeded, and 2 KMS and 1 Logging (flaky tofix)

### DIFF
--- a/spec/e2e.rb
+++ b/spec/e2e.rb
@@ -109,6 +109,7 @@ class E2E
     end
 
     def url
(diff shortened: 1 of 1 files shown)


remi:
@frankyn PTAL 😄

PR #156, opened 2017-04-05
[Cloud Storage] Adding ACL and Encryption Key Rotation samples

frankyn:
Adding the following storage related samples with minor code cleanup:

- add_bucket_owner
- remove_bucket_owner
- add_bucket_default_owner
- remove_bucket_default_owner
- print_bucket_acl
- print_bucket_acl_for_user
- add_file_owner
- remove_file_owner
- print_file_acl
- print_file_acl_for_user
- generate_signed_url
- rotate_encryption_key

@remi PTAL


storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+  puts "\tWRITER" if bucket.acl.writers.include? email
+  puts "\tREADER" if bucket.acl.readers.include? email
+  # [END print_bucket_acl_for_user]
+end
+
+def add_bucket_owner project_id:, bucket_name:, email:
+  # [START add_bucket_owner]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Your Google Cloud Storage bucket name"
+  # email       = "Google CLoud Storage ACL Entity email"
```
remi: Typo: "Google CLoud"

storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  puts "ACL for #{bucket_name}:"
+
+  bucket.acl.owners.each do |owner|
+    puts "\tOWNER #{owner}"
```
remi: Please remove the `\t` tab prefixes.  Unnecessary for snippets.

storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+  file    = bucket.file file_name
+
+  puts "ACL for #{file_name} in #{bucket_name}:"
+
+  file.acl.owners.each do |owner|
```
remi: Out of curiosity, why print only owners and readers here whereas we also print writers in some other samples above?  Do files have the same 3 roles?

storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+
+  file.acl.delete email
+
+  puts "Removed ACL permissions for #{email} from #{file_name}"
+  # [END remove_file_acl]
+end
+
+def run_sample arguments
+  command = arguments.shift
+  project_id = ENV["GOOGLE_CLOUD_PROJECT"]
```
remi: Alignment 😄

storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+
+  case command
+  when "print_bucket_acl"
+    print_bucket_acl(project_id: project_id,
```
remi: Remove parens please.

storage/acls.rb
```diff
@@ -0,0 +1,281 @@
...
+                     bucket_name: arguments.shift)
+  when "print_bucket_acl_for_user"
+    print_bucket_acl_for_user(project_id: project_id,
```
remi: Align parameters please.

storage/files.rb
```diff
@@ -257,76 +257,127 @@ def copy_file project_id:, source_bucket_name:, source_file_name:,
...
+  # current_encryption_key = "Encryption key currently being used"
+  # new_encryption_key     = "New encryption key to use"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+  file    = bucket.file file_name, encryption_key: current_encryption_key
+
+  file.rotate encryption_key: current_encryption_key,
```
remi: Minor: let's align the values of these 2 parameters with one another for consistency.

storage/files.rb
```diff
@@ -257,76 +257,127 @@ def copy_file project_id:, source_bucket_name:, source_file_name:,
...
+
+  url = file.signed_url
+
+  puts "The signed url for #{file_name} is #{url}"
+  # [END generate_signed_url]
+end
+
 def run_sample arguments
   command = arguments.shift
+  project_id = ENV["GOOGLE_CLOUD_PROJECT"]
```
remi: Alignment here for this file as well

storage/spec/acls_spec.rb
```diff
@@ -0,0 +1,435 @@
...
+  ensure
+    $stdout = real_stdout
+  end
+  attr_reader :captured_output
+
+  it "can print owner bucket acl" do
+    @bucket.acl.delete(@test_email) if email_in_bucket_acl?(@test_email)
+    @bucket.acl.reload!
+
+    expect(@bucket.acl.owners).not_to include(@test_email)
```
remi: Why do we need to delete the ACL and then re-add it before the test?

storage/spec/acls_spec.rb
```diff
@@ -0,0 +1,435 @@
...
+
+    expect(@bucket.acl.writers).not_to include(@test_email)
+
+    @bucket.acl.add_writer @test_email
+    @bucket.acl.reload!
+
+    expect(@bucket.acl.writers).to include(@test_email)
+
+    capture do
+      print_bucket_acl project_id: @project_id, bucket_name: @bucket_name
```
remi: Hmm.  I wonder if it's overkill to have multiple tests for each snippet to test all of the different possible roles?  What are your thoughts?

storage/spec/acls_spec.rb
remi: Oh, I see that this is tested using multiple different roles.  Comment continued below.

storage/spec/acls_spec.rb
```diff
@@ -0,0 +1,435 @@
...
+
+    expect(@bucket.acl.owners).not_to include(@test_email)
+
+    @bucket.acl.add_owner @test_email
+    @bucket.acl.reload!
+
+    expect(@bucket.acl.owners).to include(@test_email)
+
+    capture do
+      print_bucket_acl_for_user(project_id:  @project_id,
```
remi: Please kill parens with 🔥

storage/spec/acls_spec.rb
```diff
@@ -0,0 +1,435 @@
...
+    }.to output(
+      /Added default OWNER permission for #{@test_email} to #{@bucket_name}/
+    ).to_stdout
+
+    @bucket.default_acl.reload!
+    expect(@bucket.default_acl.owners).to include(@test_email)
+  end
+
+  it "can remove bucket default acl" do
+    if email_in_default_bucket_acl?(@test_email)
```
remi: parens 🔥

storage/spec/files_spec.rb
```diff
@@ -231,4 +238,47 @@ def capture &block
...
+      generate_signed_url(project_id:  @project_id,
+                          bucket_name: @bucket_name,
+                          file_name:   "file.txt")
+    end
+
+    expect(@captured_output).to include "The signed url for file.txt"
+
+    signed_url_from_output = @captured_output.scan(/http.*$/).first
+
+    file_contents = Net::HTTP.get URI(signed_url_from_output)
```
remi: 👍

remi:
Looks great!  I noted some parens and variable alignment and a typo, but otherwise LGTM.

storage/spec/acls_spec.rb
frankyn: My primary reason was to have coverage of the different sample conditions, OWNER, WRITER, READER. Although, I agree, it is overkill. I can remove non-owner cases. :)

storage/acls.rb
frankyn: I knew I would miss at least one alignment! :)

storage/spec/acls_spec.rb
remi: Under pretty much every other circumstance in the world, I would say let's keep them.

For code sample snippet tests, I believe the cost of maintenance and the cost of the extra LOC at the top of each of these tests outweighs the benefit.  It is preferable to have 1 example test case per snippet.  That makes it easier to navigate and understand the tests.

I love the thought.  And, in application code, all 3 branches of tests would be required.  But I say delete them in our case and have 1:1 snippet:test

remi: I do not want to set this precedent at this time.

frankyn: Does the following sound reasonable, "keep all three branches in the sample and only test the OWNER branch"?

storage/acls.rb
frankyn: Nope, only OWNER and READER is available per the gem[1] and documentation[2].
> WRITER -> N/A. You cannot apply this permission to objects.


1. https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/file/acl.rb#L77
2. https://cloud.google.com/storage/docs/access-control/lists#permissions

storage/spec/acls_spec.rb
remi: Yes.  And I would presume that the test names correlate to the snippets as well, eg. 'print bucket acl' --> it can print bucket ACLs (or something)

remi: When jumping around product samples, it's really nice to be able to quickly identify the 1 test that goes with a given snippet 😄

storage/spec/acls_spec.rb
```diff
@@ -0,0 +1,302 @@
...
+    $stdout = StringIO.new
+    block.call
+    @captured_output = $stdout.string
+  ensure
+    $stdout = real_stdout
+  end
+  attr_reader :captured_output
+
+  it "can print bucket acl" do
+    @bucket.acl.delete @test_email if email_in_bucket_acl? @test_email
```
remi: Is it still necessary to delete and re-add a bucket owner in this test?  Can we simply add owner if not already added?

frankyn: IIUC, you want to remove code from line 72 to line 76? Yea, it is unnecessary looking at it now.

remi: +1

PR #157, opened 2017-04-06
[KMS] Fix quickstart kms spec

frankyn:
The KMS quickstart lists all available key rings. The quickstart is tested by creating a new key ring, and the new key ring is expected to be included in the list of available key rings. 

The KMS quickstart test currently fails, because the list of key rings is long enough that the expected key ring is not included in the first page of available key rings.

In this PR, I added pagination support to the test, because the API doesn't support a delete key ring operation.

I'm open to other interpretations to solve this failure.

### DIFF
--- a/kms/spec/quickstart_spec.rb
+++ b/kms/spec/quickstart_spec.rb
@@ -37,7 +37,7 @@ describe "Key Management Service Quickstart" do
 
   it "can list global key rings by name" do
     test_project_id = ENV["GOOGLE_CLOUD_PROJECT"]
-    test_key_ring_id = "alist-#{test_project_id}-#{Time.now.to_i}"
+    test_key_ring_id = "list-#{test_project_id}-#{Time.now.to_i}"
     test_parent = "projects/#{test_project_id}/locations/global"
 
     test_key_ring = create_test_key_ring test_parent, test_key_ring_id
@@ -49,8 +49,21 @@ describe "Key Management Service Quickstart" do
                                          and_return(test_kms_client)
 
     expect(test_kms_client).to receive(:list_project_location_key_rings).
-                          and_wrap_original do |m, *args|
-      m.call test_parent
+        and_wrap_original do |m, *args|
+      response = m.call test_parent
+
+      key_rings = response.key_rings
+      next_page_token = response.next_page_token
+
+      while next_page_token
+        new_response = m.call test_parent, page_token: next_page_token
+
+        key_rings.concat new_response.key_rings
+
+        next_page_token = new_response.next_page_token
+      end
+
+      double({key_rings: key_rings})
     end
 
     expect {
--- a/kms/spec/quickstart_spec.rb
+++ b/kms/spec/quickstart_spec.rb
@@ -36,11 +36,12 @@ describe "Key Management Service Quickstart" do
   end
 
   it "can list global key rings by name" do
-    test_project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+    test_project_id  = ENV["GOOGLE_CLOUD_PROJECT"]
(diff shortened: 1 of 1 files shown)


kms/spec/quickstart_spec.rb
```diff
@@ -37,7 +37,7 @@ def create_test_key_ring parent, key_ring_id
 
   it "can list global key rings by name" do
     test_project_id = ENV["GOOGLE_CLOUD_PROJECT"]
-    test_key_ring_id = "alist-#{test_project_id}-#{Time.now.to_i}"
+    test_key_ring_id = "list-#{test_project_id}-#{Time.now.to_i}"
```
remi: Unrelated nit: variable assignment alignment

kms/spec/quickstart_spec.rb
```diff
@@ -49,8 +49,21 @@ def create_test_key_ring parent, key_ring_id
...
+
+      while next_page_token
+        new_response = m.call test_parent, page_token: next_page_token
+
+        key_rings.concat new_response.key_rings
+
+        next_page_token = new_response.next_page_token
+      end
+
+      double({key_rings: key_rings})
```
remi: Unrelated nit: `double key_rings: key_rings`

frankyn:
Done, thanks! @remi PTAL

remi:
Looking at what others do...

 - [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/kms/api-client/quickstart_test.py#L19) just runs the quickstart asserting that it doesn't explode
 - [.NET](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/master/kms/api/CloudKmsTest/CloudKmsTest.cs#L63) asserts that the output contains the string "keyRings/"
 - [PHP](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/kms/quickstart/test/quickstartTest.php#L43) takes the list of key rings that were pulled in the quickstart and asserts that the output includes the name of 1 of those pulled key rings
 - [Java](https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/kms/src/test/java/com/example/QuickstartIT.java#L69) asserts that the ID of a pre-created key ring appears (and it does not create additional key rings in the test)
 - [Node.js](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/kms/system-test/quickstart.test.js#L69-L72) asserts either the string "Key rings:" or "No key rings found."

If we hadn't already been polluting this project with key rings, I would vote that the test find or create a key ring by name and then test that the quickstart includes that specific key ring.

^--- basically Java

With our long list of key rings which we cannot delete, I vote...

 1. Get list of key rings.  If it's empty, then make a key ring (and get the list again).
 2. Run the quickstart
 3. Assert that the first key ring that we pulled showed up in the quickstart's output

^--- basically PHP

frankyn:
Got it, closing this PR. :)

kms/spec/quickstart_spec.rb
```diff
@@ -36,27 +45,31 @@ def create_test_key_ring parent, key_ring_id
...
+    test_key_rings   = list_test_key_rings(test_parent).key_rings
 
-    test_key_ring = create_test_key_ring test_parent, test_key_ring_id
-    expect(test_key_ring).not_to eq nil
-    expect(test_key_ring.name).to match /#{test_key_ring_id}/
+    if test_key_rings.nil?
+      test_key_ring = create_test_key_ring test_parent, test_key_ring_id
+
+      expect(test_key_ring).not_to  eq nil
+      expect(test_key_ring.name).to match /#{test_key_ring_id}/
```
remi: `.to include test_key_ring_id`

Not being used as a regular expression.

PR #158, opened 2017-04-06
Upgrade Speech to V1

remi:
Update google-cloud-speech and samples.

 - "Job" is now "Operation"
 - #recognize_job is now #process
 - language parameter is now required

See PR [Upgrade Speech to V1](https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/1380) for additional information

Also removes old Speech sample.


speech/speech_samples.rb
```diff
@@ -77,85 +83,19 @@ def speech_async_recognize_gcs project_id:, storage_path:
...
-        puts "Text: #{alternative_hypothesis['transcript']}"
-      end
-    end
-  end
-  # [END get_async_operation_results]
-end
-
-require "google/cloud/speech"
-
-if __FILE__ == $PROGRAM_NAME
```
frankyn: Why no CLI?

speech/speech_samples.rb
```diff
@@ -77,85 +83,19 @@ def speech_async_recognize_gcs project_id:, storage_path:
   require "google/cloud/speech"
 
   speech = Google::Cloud::Speech.new project: project_id
-  audio  = speech.audio storage_path, encoding: :raw, sample_rate: 16000
+  audio  = speech.audio storage_path, encoding:    :raw,
```
frankyn: Why not place both positional and named parameters in one column? I don't mind the way it looks. I'm curious :)

PR #167, opened 2017-04-14
Cloud Storage Bucket-level IAM samples

frankyn:
Bucket-level IAM Samples:

- storage_add_bucket_iam_member
- storage_view_bucket_iam_members
- storage_remove_bucket_iam_member


storage/spec/iam_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    @bucket_name = ENV["GOOGLE_CLOUD_STORAGE_BUCKET"]
+    @storage     = Google::Cloud::Storage.new
+    @project_id  = @storage.project
+    @bucket      = @storage.bucket @bucket_name
+    @test_role   = "roles/storage.admin"
+    @test_member = "user:test@test.com"
+  end
+
+  it "can view bucket IAM members" do
+    @bucket.policy do |policy|
```
remi: Whoa, what is `bucket.policy do` and why don't we show it in the samples?

storage/spec/iam_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      policy.roles[@test_role] = [@test_member]
+    end
+
+    @bucket.policy force: true
```
remi: Hmm... I can't even guess what this does.  What does this do?

storage/spec/iam_spec.rb
remi: Nevermind, we totally do!  Good 😄   Cool API

storage/spec/iam_spec.rb
remi: I wonder if we use the `force: true` param anywhere else (ie. conventionally) in the google-cloud libraries.  If not, do we use a different param?

Almost feels like `refresh: true` or `reload: true` or... `latest: true` would be better.  I can guess what any one of those do, but `force: true` didn't give me any hint of what it actually does!

remi: Non-blocking for this PR, but I'm curious of your thoughts.  Recommend suggesting different API for google-cloud-storage

frankyn: Yea, I can understand the funky nature of `force: true`. It's reloading policy roles/members. If I had to choose from the different named parameters I'd use: `reload: true`. 

What do you think about: `@bucket.policy.reload!`?

remi: `reload: true` +1

fyi @blowmage

PR #171, opened 2017-05-09
Add App Engine-specific "host" field.

p-buse:


endpoints/getting-started/openapi.yaml
```diff
@@ -6,6 +6,9 @@ info:
   version: "1.0.0"
 host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"
 # [END swagger]
+# For App Engine deployments, delete the above "host:" line and remove the "# "
```
frankyn: Hi @p-buse,

Why does the host change?

p-buse: The host changes when using App Engine, as the API is served from PROJECT-ID.appspot.com instead of *.PROJECT-ID.cloud.goog. This change is to make it easier for the user to follow the getting started guide for Endpoints on App Engine.

frankyn: Thanks @p-buse!

After reviewing the endpoints docs, I'm assuming that `host: "echo-api.endpoints.YOUR-PROJECT-ID.cloud.goog"` is not commented because it's used more often. Is that the case here?

p-buse: Yes, that's correct. The *.appspot.com is used only on App Engine, whereas the other form is used for GCE and GKE.

PR #173, opened 2017-06-15
[Video API] Adding samples

frankyn:
Adding samples for:
- quickstart
- analyze shots
- analyze labels
- analyze faces
- analyze safe search


video/quickstart.rb
```diff
@@ -0,0 +1,53 @@
...
+  annotation_result = operation.results.annotation_results.first
+
+  annotation_result.label_annotations.each do |label_annotation|
+    puts "Label description: #{label_annotation.description}"
+    puts "Locations:"
+
+    label_annotation.locations.each do |location|
+      positions = "Entire video"
+
+      if location.level != :VIDEO_LEVEL
```
remi: How about...

```ruby
locations.each do |location|
  if location.level == :VIDEO_LEVEL
    puts "Entire video"
  else
    # ... fancy things ...
    puts "#{start_offset}s thru #{end_offset}s"
  end
end
```

remi: Also, if an offset is seconds.... how about... instead of `#{offset}s` ... we do `#{start_seconds}` and put the word 'seconds' in the variable name

frankyn:
@remi PTAL

video/video_samples.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    # first result is retrieved because a single video was processed
+    annotation_result = operation.results.annotation_results.first
+
+    annotation_result.face_annotations.each do |face_annotation|
+      segment_counter = 1
+      puts "Thumbnail size: #{face_annotation.thumbnail.length}"
+
+      face_annotation.segments.each do |segment|
+        if segment.start_time_offset == -1 && segment.end_time_offset == -1
```
remi: 😱 

Do face annotations have `#locations` like the other annotations?  To check `location.level == :VIDEO_LEVEL`?

video/video_samples.rb
```diff
@@ -0,0 +1,236 @@
...
+
+  # Register a callback during the method call
+  operation = video_client.annotate_video path, features do |operation|
+    raise operation.results.message? if operation.error?
+    puts "Finished processing."
+
+    # first result is retrieved because a single video was processed
+    annotation_result = operation.results.annotation_results.first
+
+    annotation_result.face_annotations.each do |face_annotation|
```
remi: FYI if you want to know the item index when #each'ing through a collection:

```ruby
annotation_result.face_annotations.each_with_index do |face_annotation, i|
  puts "Face number #{i} is #{face_annotation}"
end
```

video/video_samples.rb
```diff
@@ -0,0 +1,236 @@
...
+      puts "Thumbnail size: #{face_annotation.thumbnail.length}"
+
+      face_annotation.segments.each do |segment|
+        if segment.start_time_offset == -1 && segment.end_time_offset == -1
+          puts "Location #{segment_counter}: Entire video"
+        else
+          start_in_seconds = segment.start_time_offset / 1000000.0
+          end_in_seconds   = segment.end_time_offset / 1000000.0
+
+          puts "Location #{segment_counter}: #{start_in_seconds} through #{end_in_seconds}"
```
remi: Can we `puts "#{start_in_seconds} through #{end_in_seconds}"` like we do with the other samples?  It's not clear to me what the value of the segment_counter is - I'd love to exclude it from the sample.

video/video_samples.rb
```diff
@@ -0,0 +1,236 @@
...
+    puts "Finished processing."
+
+    # first result is retrieved because a single video was processed
+    annotation_result = operation.results.annotation_results.first
+
+    annotation_result.safe_search_annotations.each do |safe_search_annotation|
+      time_in_seconds = safe_search_annotation.time_offset / 1000000.0
+
+      puts "Time: #{time_in_seconds}"
+      puts "\tadult:   #{safe_search_annotation.adult}"
```
remi: Who is tracy?

Hard to read the words 😕 

```ruby
puts "\tadult:   #{safe_search_annotation.adult}"
puts "\tspoof:   #{safe_search_annotation.spoof}"
puts "\tmedical: #{safe_search_annotation.medical}"
puts "\tracy:    #{safe_search_annotation.racy}"
puts "\tviolent: #{safe_search_annotation.violent}"
```

Less challenging to read the words 😄 

```ruby
puts "adult:   #{safe_search_annotation.adult}"
puts "spoof:   #{safe_search_annotation.spoof}"
puts "medical: #{safe_search_annotation.medical}"
puts "racy:    #{safe_search_annotation.racy}"
puts "violent: #{safe_search_annotation.violent}"
```

video/video_samples.rb
```diff
@@ -0,0 +1,236 @@
...
+
+    # first result is retrieved because a single video was processed
+    annotation_result = operation.results.annotation_results.first
+    shot_counter      = 1
+
+    annotation_result.shot_annotations.each do |shot_annotation|
+      start_in_seconds = shot_annotation.start_time_offset / 1000000.0
+      end_in_seconds   = shot_annotation.end_time_offset / 1000000.0
+
+      puts "Scene #{shot_counter}: #{start_in_seconds} to #{end_in_seconds}"
```
remi: I love the variable name: shot_counter

But what do you think about this for consistency and simplicity:

```ruby
puts "#{start_in_seconds} through #{end_in_seconds}"
```

Users can increment numbers when looping on their own time 😋 

Or, if you want...

```ruby
puts "Scene: #{start_in_seconds} to #{end_in_seconds}"
```

^--- if we do this, consider whether it would be useful to update the others in a similar manner

video/video_samples.rb
frankyn: This is the best comment I've ever read lol Who is Tracy? I don't know hahaha

video/video_samples.rb
frankyn: No it does not. :(

video/spec/video_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      local_tempfile.close
+      local_tempfile.unlink
+    end
+  end
+
+  it "can analyze faces from a gcs file" do
+    expect {
+      analyze_faces path: "gs://#{@labels_file}"
+    }.to output(
+      /Thumbnail size: \d+/
```
remi: Presuming it doesn't break, will the Video API always return the same value for Thumbnail size here?

 - The tests here for Labels and Safe Search use hardcoded values which the API is expected to return.
 - The tests here for Faces and Shot Change Detection use regex matches

If it's simple to add explicit values for Faces and Shot Change Detection which shouldn't change, maybe use those?  Else I'm +1 for the current matchers!

video/video_samples.rb
remi: Oh noooeeessss....

Hmm...

remi: It's a *little bit* more clear without the counter at least?

```ruby
operation = video_client.annotate_video path, features do |operation|
    raise operation.results.message? if operation.error?
    puts "Finished processing."

    # first result is retrieved because a single video was processed
    annotation_result = operation.results.annotation_results.first

    annotation_result.face_annotations.each do |face_annotation|
      puts "Thumbnail size: #{face_annotation.thumbnail.length}"

      face_annotation.segments.each do |segment|
        if segment.start_time_offset == -1 && segment.end_time_offset == -1
          puts "Entire video"
        else
          start_in_seconds = segment.start_time_offset / 1000000.0
          end_in_seconds   = segment.end_time_offset / 1000000.0

          puts "#{start_in_seconds} to #{end_in_seconds}"
        end
      end
    end
  end
```

I can't think of a way to get rid of or clean up this horribly ugliness tho:

```ruby
if segment.start_time_offset == -1 && segment.end_time_offset == -1
```

Take it as is it is for now... 😕

remi: I was trying to look at the proto returned by face annotations... and I noticed that `FACE_DETECTION` isn't actually a feature?

https://cloud.google.com/video-intelligence/docs/reference/rpc/google.cloud.videointelligence.v1beta1#feature

??? I'm confused

@gguuss ?

frankyn: synced with remi offline

frankyn:
@remi PTAL Thank you!

PR #175, opened 2017-06-27
[Spanner] Add Quickstart

frankyn:
Adding Spanner quickstart


spanner/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Install Dependencies
+
+1. Install the [Bundler](http://bundler.io/) gem.
+
+1. Install dependencies using:
+
+    `bundle install`
+
+## Run Quickstart
```
remi: Add something about setting the 2 env vars for instance and database 🤘

remi: Nm! We can think about adding how to run the tests to READMEs late on! But this LGTM 🤘

PR #176, opened 2017-07-05
[Auth] Add Ruby Client Library Auth samples

frankyn:
This PR is to fill Ruby samples for https://cloud.google.com/docs/authentication/production.


auth/README.md
```diff
@@ -0,0 +1,8 @@
+<img src="https://avatars2.githubusercontent.com/u/2810941?v=3&s=96" alt="Google Cloud Platform logo" title="Google Cloud Platform" align="right" height="96" width="96"/>
+
+# Google Cloud Ruby Client Library Auth Samples
+
+The samples show different authentication workflows using the
+[Google Cloud Ruby Client Library][ruby_client].
+
+[ruby_client]: https://github.com/GoogleCloudPlatform/google-cloud-rub://googlecloudplatform.github.io/google-cloud-ruby/#
```
remi: Copy paste typo 😸

auth/auth.rb
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+
+def implicit project_id:
+  # [START implicit]
+  # project_id = "Your Google Cloud project ID"
+
+  require "google/cloud/storage"
+
+  # If you don't specify credentials when constructing the client, the client
+  # library will look for credentials in the environment.
```
remi: **User:** *"look for credentials in the environment"? What does that mean?*

It's a little vague. Is that the same wording that the other samples use? If not, maybe see if they have better language. It's very specific, but I think it would actually be very beneficial to users to specify *exactly* where we look for credentials, eg. say the library will look for credentials from the X environment variable? If there's a clear way to describe the specifics, I would *love* to know the exact environment variable / etc as a user. I mean... that's why I'm on the page that described the Authentication details in the first place! 😄

frankyn: Understood, follow-up response:

* This comment is used by other language samples ([Python](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/auth/cloud-client/snippets.py#L24) and [Node](https://github.com/GoogleCloudPlatform/nodejs-docs-samples/blob/master/auth/auth.js#L28)).

* The environment variable is documented on the page https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application. I didn't want to repeat the same wording in the sample to reflect a similar sample used by other languages.

remi: Ohhh... the env var is on the page *right* before the code snippet, and it's `highlighted as monospace and easy to see` Cool deal 👍

PR #178, opened 2017-07-07
Added Proto and JSON examples

palladius:
Added (in comment) a JSON and Protobuf example.
- JSON is just reaqdy to go. And tested.
- Protobuf is a hint on how to do it - could use a better example.


frankyn:
Closing a year old.

PR #183, opened 2017-07-31
[Speech] Demonstrate Word Time Offsets feature of Async Recognize

remi:
Updates google-cloud-speech to 0.26.0 which supports this option


speech/spec/speech_samples_spec.rb
```diff
@@ -64,12 +75,19 @@
...
       speech_async_recognize_gcs project_id:   @project_id,
                                  storage_path: path
-    }.to output(
-      "Operation started\nTranscription: how old is the Brooklyn Bridge\n"
-    ).to_stdout
+    end
+
+    expect(captured_output).to include "Operation started"
+    expect(captured_output).to include "Transcription: how old is the Brooklyn Bridge"
+    expect(captured_output).to include "Word: how 0 0.3"
```
remi: FYI I went ahead and included the exact value.

If the start/end time of the words in this audio change... something's probably very wrong! 😄

frankyn: Thanks, I would have asked about this

speech/Gemfile.lock
```diff
@@ -98,4 +98,4 @@ DEPENDENCIES
   rspec
 
 BUNDLED WITH
-   1.14.6
+   1.15.2
```
frankyn: Nit: Bundler version is at 1.15.3.

PR #185, opened 2017-08-02
[Storage] Requester pays samples

frankyn:
This pr contains the following samples:

* download_file_requester_pays
* get_requester_pays_status
* enable_requester_pays
* disable_requester_pays


frankyn:
@remi PTAL

storage/spec/files_spec.rb
```diff
@@ -182,6 +185,94 @@ def capture &block
...
+      }.to output(
+        "Downloaded file.txt using billing project #{@project_id_secondary}\n"
+      ).to_stdout
+
+      expect(File.size local_file.path).to be > 0
+      expect(File.read local_file.path).to eq(
+        "Content of test file.txt\n"
+      )
+
+      @bucket.policy do |policy|
```
remi: Is this cleanup code? 

If storage specs need the bucket policy to be set in a certain way, let's override the bucket policy in the before block 

Cleanup code should be in the after block - because the test could blow up on any line! If it fails, the cleanup code will never run 😄

remi:
LGTM, Test Master Extraordinaire!

PR #191, opened 2017-08-18
KMS: Make the samples consistent with the other languages.

russ-:
In particular, this change:

- changes to use a consistent "_id" suffix for resource name components
- spells "crypto key" and "key ring" as separate words
- orders location between project and key ring, as in the KMS resource hierarchy
- uses "plaintext" and "ciphertext" consistently to refer to unecrypted and encrypted data, respectively
- fixes some of the text that mixes up resource IDs and resource names

I also updated the quickstart test to work with projects that already have key rings, since it took me a bit to understand why it wasn't passing for me.

Tracking bug: http://b/64758639

### DIFF
--- a/kms/kms.rb
+++ b/kms/kms.rb
@@ -510,113 +510,113 @@ def run_sample arguments
   case command
   when "create_key_ring"
     $create_key_ring.call(
-      project_id: project_id,
-      location_id: arguments.shift,
-      key_ring_id: arguments.shift
+      project_id:    project_id,
+      location_id:   arguments.shift,
+      key_ring_id:   arguments.shift
     )
   when "create_crypto_key"
     $create_crypto_key.call(
-      project_id: project_id,
-      location_id: arguments.shift,
-      key_ring_id: arguments.shift,
+      project_id:    project_id,
+      location_id:   arguments.shift,
+      key_ring_id:   arguments.shift,
       crypto_key_id: arguments.shift
     )
   when "encrypt_file"
     $encrypt.call(
-      project_id: project_id,
-      location_id: arguments.shift,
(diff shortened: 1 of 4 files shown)


kms/spec/quickstart_spec.rb
```diff
@@ -46,11 +46,14 @@ def list_test_key_rings parent
...
   it "can list global key rings by name" do
     test_project_id  = ENV["GOOGLE_CLOUD_PROJECT"]
-    test_key_ring_id = "a-keyring-list-#{test_project_id}"
+    test_key_ring_id = "a-key-ring-list-#{test_project_id}"
     test_parent      = "projects/#{test_project_id}/locations/global"
     test_key_rings   = list_test_key_rings(test_parent).key_rings
 
-    if test_key_rings.nil?
+    created = test_key_rings.any? do
+      |key_ring| key_ring.name.end_with?(test_key_ring_id)
```
frankyn: Move `|key_ring|` to be after `do`.

kms/kms.rb
```diff
@@ -506,135 +508,135 @@ def run_sample arguments
...
-      location: arguments.shift
+      location_id: arguments.shift,
+      key_ring_id: arguments.shift
     )
-  when "create_cryptokey"
-    $create_cryptokey.call(
+  when "create_crypto_key"
+    $create_crypto_key.call(
       project_id: project_id,
+      location_id: arguments.shift,
```
russ-: Done.

kms/spec/quickstart_spec.rb
russ-: Done.

frankyn:
Thanks @russ- merging now.

PR #193, opened 2017-08-30
[Storage] Add samples for labels and create bucket with location and storage_class

frankyn:
This PR is to support the following samples:

* get_bucket_labels
* add_bucket_label
* remove_bucket_label
* create_bucket_class_location (create bucket with location and storage_class).

Also, fixed a few issues that made it through in the requester pays PR.

### DIFF
--- a/storage/buckets.rb
+++ b/storage/buckets.rb
@@ -116,7 +116,7 @@ def list_bucket_labels project_id:, bucket_name:
   storage = Google::Cloud::Storage.new project: project_id
   bucket  = storage.bucket bucket_name
 
-  puts "Updated labels for #{bucket_name}"
+  puts "Labels for #{bucket_name}"
   bucket.labels.each do |key, value|
     puts "#{key} = #{value}"
   end
--- a/storage/buckets.rb
+++ b/storage/buckets.rb
@@ -90,6 +90,68 @@ def create_bucket project_id:, bucket_name:
   # [END create_bucket]
 end
 
+def get_bucket_labels project_id:, bucket_name:
+  # [START get_bucket_labels]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Name of your Google Cloud Storage bucket"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  puts "Updated labels for #{bucket_name}"
+  bucket.labels.each do |key, value|
+    puts "#{key} = #{value}"
+  end
+  # [END get_bucket_labels]
+end
+
+def add_bucket_label project_id:, bucket_name:, label_key:, label_value:
+  # [START add_bucket_label]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Name of your Google Cloud Storage bucket"
+  # label_key   = "Cloud Storage bucket Label Key"
+  # label_value = "Cloud Storage bucket Label Value"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  bucket.update do |bucket|
+    bucket.labels[label_key] = label_value
+  end
+
+  puts "Updated labels for #{bucket_name}"
+  bucket.labels.each do |key, value|
+    puts "#{key} = #{value}"
+  end
+  # [END add_bucket_label]
+end
+
+def remove_bucket_label project_id:, bucket_name:, label_key:
+  # [START remove_bucket_label]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Name of your Google Cloud Storage bucket"
+  # label_key   = "Cloud Storage bucket Label Key"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  bucket.update do |bucket|
+    bucket.labels[label_key] = nil
+  end
+
+  puts "Updated labels for #{bucket_name}"
+  bucket.labels.each do |key, value|
+    puts "#{key} = #{value}"
+  end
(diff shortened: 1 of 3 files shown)


storage/buckets.rb
```diff
@@ -90,6 +90,86 @@ def create_bucket project_id:, bucket_name:
   # [END create_bucket]
 end
 
+def create_bucket_class_location project_id:, bucket_name:, location:, storage_class:
+  # [START create_bucket_class_location]
+  # project_id    = "Your Google Cloud project ID"
+  # bucket_name   = "Name of Google Cloud Storage bucket to create"
+  # location      = "Location of your Google Cloud Storage bucket"
+  # storage_class = "Storage class of your Google Cloud Storage bucket"
```
remi: What is an example of what this looks like?

I wonder if it would be more clear if `location` and `storage_class` were explicit rather than variables.

For example of explicit options, see Speech Recognize:

```ruby
audio = speech.audio audio_file_path, encoding:    :linear16,   # <--- All Hard Coded Values
                                      sample_rate: 16000,       #
                                      language:    "en-US"      #
```

I know that this sample will appear within the documentation which should mention the available values, but I can't personally even *guess* what a correct value for `storage_class`might be

storage/buckets.rb
```diff
@@ -90,6 +90,86 @@ def create_bucket project_id:, bucket_name:
...
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.create_bucket bucket_name,
+                                  location:      location,
+                                  storage_class: storage_class
+
+  puts "Created bucket #{bucket.name} in #{location} with storage class #{storage_class}"
+  # [END create_bucket_class_location]
+end
+
+def get_bucket_labels project_id:, bucket_name:
```
remi: `get_bucket_labels` --> `list_bucket_labels`

For CRUD types of samples, we name them `get_single_resource` or `list_resources` like `get_bucket` and `list_buckets`

If this conflicts with our defined standard doc region tags, then this is OK. I just don't like it 😝

storage/buckets.rb
```diff
@@ -90,6 +90,86 @@ def create_bucket project_id:, bucket_name:
...
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  bucket.update do |bucket|
+    bucket.labels[label_key] = label_value
+  end
+
+  puts "Updated labels for #{bucket_name}"
```
remi: Why list the labels here? The sample demonstrates *adding*

Can we just `puts "Added label to bucket"?

storage/buckets.rb
```diff
@@ -90,6 +90,86 @@ def create_bucket project_id:, bucket_name:
...
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  bucket.update do |bucket|
+    bucket.labels[label_key] = nil
+  end
+
+  puts "Updated labels for #{bucket_name}"
+  bucket.labels.each do |key, value|
```
remi: Same comment as above. Remove listing labels after add/remove operations.

storage/buckets.rb
frankyn: Understood, will update to use explicit values instead. I got used to variable, and the canonical uses explicit values.

frankyn:
I updated the PR based on review. I didn't change region tags because they're being used by canonical sample. I'll be more careful next time.

storage/buckets.rb
```diff
@@ -90,6 +90,78 @@ def create_bucket project_id:, bucket_name:
...
+  # [START get_bucket_labels]
+  # project_id  = "Your Google Cloud project ID"
+  # bucket_name = "Name of your Google Cloud Storage bucket"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  puts "Updated labels for #{bucket_name}"
```
remi: Typo. This doesn't Update :)

frankyn:
Thanks @remi! Typo is now fixed!

storage/buckets.rb
```diff
@@ -90,6 +90,78 @@ def create_bucket project_id:, bucket_name:
...
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new project: project_id
+  bucket  = storage.bucket bucket_name
+
+  bucket.update do |bucket|
+    bucket.labels[label_key] = label_value
+  end
+
+  puts "Added label to #{bucket_name}"
```
remi: Reading back through this... now I'm thinking that having the output include the key and value would be nice and clear. I think I argued to make this simpler, but it's maybe slightly too simple.

`puts "Added label #{label_key} with value #{label_value} to #{bucket_name}"`

Thoughts?

frankyn: As a user it'd nice to print key and value, but a user could also list labels for a bucket after making a new label to verify.

remi: Per previous comments, I really don't think it makes sense to **list** the labels in an **add** or **remove** sample.

Listing samples is substantial enough of a task that it has *its own sample snippet*

For samples, we print information about the operation. We do not **prove** that it works.

What we [create a Spanner database](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/f2b32b69406a7c8af36bf9f5f06441c380ac9260/spanner/spanner_samples.rb#L46) we don't print out all Spanner databases to prove that the operation was successful.

remi: But when we create a Spanner database, we do show the name of the database we added and what instance

```ruby
puts "Created database #{database_id} on instance #{instance_id}"
```

That's why I think it would be appropriate to include the key and value.

It's more comprehensive IMO and consistent with other samples.

frankyn: Gotcha, I'll update the sample to include key and value. Consistency!

remi: 🤘

frankyn:
Thanks @remi I have updated the PR.

PR #194, opened 2017-09-03
[Stackdriver] Samples for documentation. 

frankyn:
This PR is to provide samples for the Stackdriver documentation.


stackdriver/README.md
```diff
@@ -0,0 +1,50 @@
+# Stackdriver Samples
+
+Samples used by Stackdriver documentation. The samples don't work out-of-the-box
```
remi: We don't need to point out that *"the samples don't work out-of-the-box"*. Let's remove that sentence.

stackdriver/README.md
```diff
@@ -0,0 +1,50 @@
...
+and require some modification. The samples are provided as reference guides
+when using Rails, Sinatra, other-rack frameworks, and non-rack apps.
```
remi: > *"The samples are provided as reference guides for using Stackdriver with Rails, Sinatra, Rack, and pure Ruby"*

stackdriver/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Rails
+
+The `rails_configuration.rb` provides a configuration example that can be used
+in your `config/environments/*.rb` files.
+
+## Sinatra
+
+Sinatra exmple apps load middleware using a rack configuration file. For example
```
remi: Type: `exmple`

stackdriver/logging.ru
```diff
@@ -0,0 +1,20 @@
+require_relative "sinatra_logging"
```
remi: Using rackup files for Sinatra examples adds confusion.

See [Sinatra: README](http://www.sinatrarb.com/intro.html). The intro uses a simple `myapp.rb` file.

In my experience with Sinatra, I *always* start with a simple .rb file. That is idiomatic Sinatra. Frequently `app.rb` which is run simple via `ruby app.rb`

Let's only use a rackup file for the Rack examples.

stackdriver/non_rack_debugger.rb
```diff
@@ -0,0 +1,13 @@
+# [START explicit_debugger_norack]
```
remi: `non_rack_debugger.rb` --> `ruby_debugger.rb`

Let's not "non-rack". It's simply pure Ruby 😄

stackdriver/sinatra_error_reporting.rb
```diff
@@ -0,0 +1,18 @@
+require "sinatra"
+
+get "/" do
+
```
remi: This is missing a response. At least respond with something

```ruby
get "/" do
  erb :template
end
```

remi: Actually, another example appears to use "hello world" so use that or make them consistent

stackdriver/Gemfile
```diff
@@ -0,0 +1,7 @@
+source "https://rubygems.org"
+
+gem "sinatra", "~> 1.4"
```
remi: Y?

stackdriver/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Rails
+
+The `rails_configuration.rb` provides configuration examples that can be used
+in your `config/environments/*.rb` files.
+
+## Rack and Sinatra
+
+Sinatra example apps load middleware using a rack configuration file. Sinatra
+samples are prefixed with `sinatra_` and Rack examples use the extension `.ru`.
```
remi: This is so clear! README is really great 👍

stackdriver/error_reporting.ru
```diff
@@ -0,0 +1,17 @@
...
+end
+# [END error_reporting_configure]
+
+# [START error_reporting_middleware]
+require "google/cloud/error_reporting"
+
+use Google::Cloud::ErrorReporting::Middleware
+# [END error_reporting_middleware]
+
+run Proc.new { |env| ["200", {"Content-Type" => "text/html"}, ["Hello world!"]] }
```
remi: Most folks would use the dash rocket, `run -> { |env| ... }`

stackdriver/Gemfile
frankyn: The integrations tests only cover 1.4 at the moment. I requested that Heng update google-cloud-ruby tests to include Sinatra 2.0 as it makes sense moving forward. 

Current tested versions:
https://github.com/GoogleCloudPlatform/google-cloud-ruby/tree/master/integration

Cur

frankyn:
Thanks @remi I have updated the PR!

stackdriver/Gemfile
frankyn: Tested the samples manually with Sinatra 2.0

PR #195, opened 2017-09-05
Print out all the speech results

jerjou:
There was some confusion where folks thought the API was only transcribing the first portion of the audio, when in fact it was because they were only using the first result. I'm updating the samples to try to reduce that confusion. See b/64934785

### DIFF
--- a/speech/speech_samples.rb
+++ b/speech/speech_samples.rb
@@ -25,9 +25,10 @@ def speech_sync_recognize project_id: nil, audio_file_path: nil
                                          language:    "en-US"
 
   results = audio.recognize
-  result  = results.first
 
-  puts "Transcription: #{result.transcript}"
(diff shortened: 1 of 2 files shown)


speech/speech_samples.rb
```diff
@@ -25,9 +25,11 @@ def speech_sync_recognize project_id: nil, audio_file_path: nil
                                          language:    "en-US"
 
   results = audio.recognize
-  result  = results.first
 
-  puts "Transcription: #{result.transcript}"
+  # Each result represents a consecutive portion of the audio
```
remi: Remove this comment for each change in `speech_samples.rb`

The quickstart is heavily commented but not the documentation snippets that come from this file

jerjou: Done.

PR #199, opened 2017-09-18
[DO NOT MERGE] [Video] Library Interface Review

remi:


video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
+# DO NOT MERGE - this file for comments only
+
+def analyze_labels_gcs path:
+  # [START analyze_labels_gcs]
+  # path = "Path to a video file on Google Cloud Storage: gs://bucket/video.mp4"
+
+  require "google/cloud/video_intelligence/v1beta1"
+
+  video_client = Google::Cloud::VideoIntelligence::V1beta1::VideoIntelligenceServiceClient.new
```
remi: Obvs the require and client lib instantiation will look good with regen 😄

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+  features     = [:LABEL_DETECTION]
```
remi: OOC: will lower-case symbols work today, eg. would `:label_detection` work?

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+
+  # Register a callback during the method call
+  operation = video_client.annotate_video path, features do |operation|
+    raise operation.results.message? if operation.error?
```
remi: What is the value of `message` for *successful* responses?

Is this used **only** for errors?  If so, can we call it `error_message`?

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+    puts "Finished Processing."
+
+    # first result is retrieved because a single video was processed
+    annotation_result = operation.results.annotation_results.first
```
remi: Hmm. What attributes does a Result object have besides `annotation_results`??

This is weird. I'd figure the `operation.results` would already be an array of `AnnotationResult` or some such

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+
+    annotation_result.label_annotations.each do |label_annotation|
+      puts "Label description: #{label_annotation.description}"
+      puts "Locations:"
+
+      label_annotation.locations.each do |location|
+        if location.level == :VIDEO_LEVEL
```
remi: This is silly IMO. Dunno what we can do about it easily. But... really, a Ruby developer should call a predicate method to check if the "location" respresents the whole video

```ruby
if location.entire_video?
  # ...
else
  # ...
end
```

video/video_library_interface_review.rb
geigerj: That's the only field; it's defined in the proto here: https://github.com/googleapis/googleapis/blob/master/google/cloud/videointelligence/v1beta2/video_intelligence.proto#L310

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+  require "base64"
+  require "google/cloud/video_intelligence/v1beta1"
+
+  video_client  = Google::Cloud::VideoIntelligence::V1beta1::VideoIntelligenceServiceClient.new
+  features      = [:LABEL_DETECTION]
+  video_file    = File.read path
+  encoded_video = Base64.encode64 video_file
+
+  # Register a callback during the method call
+  operation = video_client.annotate_video nil, features, input_content: encoded_video do |operation|
```
remi: WTF?

`annotate_video nil`

Why in the world.... is the purpose of this sample to display how someone can pass BYTES instead of a PATH?

Maybe this is a sample problem. There could be 1 sample that shows people how to use BYTES. But this sample (for labels) seems to randomly not use a PATH? Why does the first sample use a PATH but this uses BYTES?

We obviously would like to get away from users having to READ and ENCODE64 files themselves if the library can handle a path 😄 
^--- this feedback comes back often from DPEs

remi: ## BLOCKER!

That `nil` argument needs to go away 😛

remi: If you want a named param for this, then...

With GCS path

```ruby
operation = video_client.annotate_video features, gcs_path: "gcs://foo"
```

```ruby
operation = video_client.annotate_video features, video_content: encoded_video
```

NOTE: The AnnotateVideoRequest Proto should have a ONE OF for input_uri vs input_content

OK... looking at the proto, I think `input_uri` and `input_content` are satisfactory and clear 😄

geigerj: TODO: short-term -- we can mark `input_uri` optional in GAPIC config. longer-term -- the API should maybe consider using a oneof for `{input_uri, input_content}`?

video/video_library_interface_review.rb
geigerj: Not currently. This would be a straightforward feature to add in GAX, though.

video/video_library_interface_review.rb
```diff
@@ -0,0 +1,184 @@
...
+  # [START analyze_labels_gcs]
+  # path = "Path to a video file on Google Cloud Storage: gs://bucket/video.mp4"
+
+  require "google/cloud/video_intelligence/v1beta1"
+
+  video_client = Google::Cloud::VideoIntelligence::V1beta1::VideoIntelligenceServiceClient.new
+  features     = [:LABEL_DETECTION]
+
+  # Register a callback during the method call
+  operation = video_client.annotate_video path, features do |operation|
```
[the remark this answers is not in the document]
geigerj: Not currently. This would be a straightforward feature to add in GAX, though.

video/video_library_interface_review.rb
frankyn: Chiming in:  You can define wild card gcs:// paths for multiple videos as noted in [AnnotateVideoRequest](https://cloud.google.com/video-intelligence/docs/reference/rpc/google.cloud.videointelligence.v1#annotatevideorequest)

frankyn:
Closing PR as samples were updated to use Video v1 API. Please, reopen if this PR is still necessary.

PR #200, opened 2017-09-18
[Speech] gapic Speech samples

frankyn:
Not complete, pending an issue with Ruby similar to Python issue resolved by [toolkit#1571](https://github.com/googleapis/toolkit/pull/1571).

cc: @geigerj


speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
 # See the License for the specific language governing permissions and
```
remi: This file shows the samples :)

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
+  speech = Google::Cloud::Speech.new
 
-  results = audio.recognize
+  audio_file = File.binread audio_file_path
+  config     = { encoding: :LINEAR16, sample_rate_hertz: 16000, language_code: "en-US" }
+  audio      = { content: audio_file }
 
-  results.each do |result|
-    puts "Transcription: #{result.transcript}"
+  response = speech.recognize config, audio
```
remi: # Ruby Idiom Issue 🛑

remi: With Ruby, when methods accept a hash... Rubyists pass the hash parameter in a way such that it LOOKS like normal named parameters. Just by reading the code, there actually is no way to tell if the method takes a Hash or regular named params

remi: SOLUTION: Instead of `recognize` taking 2 params, it should take 1

That way, people can...

```ruby
recognize content: audio_file, encoding: :LINEAR16, sample_rate_hertz: 16000, language_code: "en-US"
```

Currently, you have to do this ☹️ 

```ruby
recognize({ content: audio_file }, encoding: :LINEAR16, sample_rate_hertz: 16000, language_code: "en-US")
```

NOTE: parens **required** ☹️

remi: **IDEAL** Can we have `content` be the first positional argument again? 😄

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
+
+  alternatives = response.results.first.alternatives
```
remi: What is `results` if it's not already an array of `alternatives`?

*...need to look at recognize result...*

Not great looking... wondering...

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
 
-    result.words.each do |word|
-      puts "Word: #{word.word} #{word.start_time} #{word.end_time}"
+  alternatives.each do |alternative|
+    puts "Transcription: #{alternative.transcript}"
+
+    alternative.words.each do |word|
+      start_time = word.start_time.seconds + word.start_time.nanos/1000000000.0
```
remi: Yuck.

@frankyn What type/value is `word.start_time` in GAPIC only? Versus what was the type/value of `word.start_time` in hand-written?

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
 
-  speech = Google::Cloud::Speech.new project: project_id
-  audio  = speech.audio storage_path, encoding:    :linear16,
-                                      sample_rate: 16000,
-                                      language:    "en-US"
+  speech = Google::Cloud::Speech.new
 
-  operation = audio.process
+  config = { encoding: :LINEAR16, sample_rate_hertz: 16000, language_code: "en-US" }
+  audio  = { uri: storage_path }
```
remi: # ⚠️  Video uses `input_uri` and `input_content` but Speech uses `content` or `uri`

speech/speech_samples.rb
remi: ## ❗️ ACTUALLY.... 

`content:` or `uri:` don't fit in very well with `encoding:` `sample_rate_hertz:` etc...

And Video uses different `input_content:` and `input_uri:`

Could the named params be consistent???

I'm OK with the named params being all grouped together ... but let's make them consistent ...

### Let's put a new method on top of this and handle the params ourselves... `**kwargs` and handle input_uri and input_content

remi:
```ruby
def recognize **kwargs
  audio = {}
  if kwargs[:audio_uri]
    audio[:uri] = kwargs.delete :audio_uri
  elsif
    audio[:content] = kwargs.delete :audio_content
  else
    raise "No dawg, u needs audio yo! 🐕"
  end

  super audio, kwargs
end
```

remi: #### *Consistency please too* 😄

speech/speech_samples.rb
remi: ### ⚠️ FRICTION yo

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
+  config     = { encoding: :LINEAR16, sample_rate_hertz: 16000, language_code: "en-US" }
+  audio      = { content: audio_file }
 
-  puts "Operation started"
+  operation = speech.long_running_recognize config, audio
 
   operation.wait_until_done!
 
-  results = operation.results
+  raise operation.results.message if operation.error?
```
remi: What is in `results.message` on success?

Is it only used for error messages? If so, why not call it `error_message`?

speech/speech_samples.rb
remi: Could we get some more details about the changes here? 😄

speech/speech_samples.rb
remi: If `response` only has `results` ... can `response` BE `results`? Maybe? Please? 😺

speech/speech_samples.rb
```diff
@@ -12,160 +12,168 @@
...
   require "google/cloud/speech"
 
-  speech = Google::Cloud::Speech.new project: project_id
-  audio  = speech.audio audio_file_path, encoding:    :linear16,
-                                         sample_rate: 16000,
-                                         language:    "en-US"
+  speech = Google::Cloud::Speech.new
 
-  operation = audio.process
+  audio_file = File.binread audio_file_path
```
remi: It would of course be dope if we could accept a local file path 😄 

For example y'all ...

```ruby
def recognize **kwargs
  audio = {}
  if kwargs[:audio_uri]
    audio[:uri] = kwargs.delete :audio_uri
  elsif kwargs[:audio_content]
    audio[:content] = kwargs.delete :audio_content
  elsif kwargs[:audio_path]
    audio[:content] = File.binread kwargs.delete(:audio_path)
  else
    raise "No dawg, u needs audio yo! 🐕"
  end

  super audio, kwargs
end
```

remi: Hey hey @frankyn et al ... Look! An idea for Path! How easy, amiright??

@jgeiger @lukesneeringer

remi: # 😇

speech/speech_samples.rb
```diff
@@ -12,210 +12,314 @@
...
+
+  alternatives = operation.response.results.first.alternatives
+
+  alternatives.each do |alternative|
+    puts "Transcription: #{alternative.transcript}"
 
-  results.each do |result|
-    puts "Transcription: #{result.transcript}"
+    alternative.words.each do |word|
+      start_time = word.start_time.seconds + word.start_time.nanos/1000000000.0
```
remi: This makes my very very sad

PR #202, opened 2017-09-27
[Storage] Canonical Set Metadata sample

frankyn:
This PR is to add sample for setting fixed key and custom metadata for a Cloud Storage file.


frankyn:
Updated PR to fix request. Merging based on conversation from last week of fixing and then able to merge.

PR #205, opened 2017-10-19
[Spanner] Add Read Stale Data sample

frankyn:
This PR is to add a Read Stale Data sample for Ruby.

@remi PTAL, if you're unable to review by this Friday, I'll review it as you recommended.


frankyn:
Did a final pass and will now merge.

PR #214, opened 2017-11-16
Keeping costs low

devlance:
We've gotten reports of unexpected costs from users that left the hello world running and had two standard instances running for a month. This keeps the costs to a minimum to kick the tires.

### DIFF
--- a/appengine/hello_world/app.yaml
+++ b/appengine/hello_world/app.yaml
@@ -16,4 +16,16 @@
 runtime: ruby
 env: flex
 entrypoint: bundle exec ruby app.rb
+
+# This sample incurs costs to run on the App Engine flexible environment. 
+# The settings below are to reduce costs during testing and are not appropriate
+# for production use. For more information, see:
+# https://cloud.google.com/appengine/docs/flexible/python/configuring-your-app-with-app-yaml
+manual_scaling:
+  instances: 1
+resources:
+  cpu: 1
+  memory_gb: 0.5
+  disk_size_gb: 10
+
 # [END app_yaml]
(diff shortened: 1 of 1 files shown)


appengine/hello_world/app.yaml
```diff
@@ -16,4 +16,16 @@
 runtime: ruby
 env: flex
 entrypoint: bundle exec ruby app.rb
+
```
frankyn: Did you mean to put in within the region tag? If not can you move out of the resource and manual scaling configuration out of the region tags for simplicity.

devlance: Yes, I want this to be part of the config that is pulled into the quickstart docs.

appengine/hello_world/app.yaml
```diff
@@ -16,4 +16,16 @@
...
+# This sample incurs costs to run on the App Engine flexible environment. 
+# The settings below are to reduce costs during testing and are not appropriate
+# for production use. For more information, see:
+# https://cloud.google.com/appengine/docs/flexible/python/configuring-your-app-with-app-yaml
```
frankyn: https://cloud.google.com/appengine/docs/flexible/ruby/configuring-your-app-with-app-yaml

appengine/hello_world/app.yaml
frankyn: FWIW, documentation should provide the description below and not be in the app.yaml. In case it needs to be changed it can be updated directly in the docs instead of the app.yaml.

devlance: Actually, I think it makes more sense to have it embedded within the code than within the docs. The code and the side affect of running it could be the same regardless of if your discovery point is the docs. We don't want to surprise users with unexpected bills regardless of how they come across the sample.

frankyn: Understood, thanks for sharing your perspective.

PR #215, opened 2017-11-29
[Spanner] Use bind parameters in query_data_with_index

frankyn:
This is a PR to update the `query_data_with_index` sample to use bind parameters.


frankyn:
Thanks @jsimonweb for finding that! Update PTAL.

PR #222, opened 2017-12-14
Update deployment.yaml

navinger:
For better readability, let's use long name options for the ESP config. The doc is here: https://cloud.google.com/endpoints/docs/openapi/specify-proxy-startup-options


frankyn:
Thanks Paul! I missed this PR.

PR #228, opened 2018-01-16
[Storage] Add public data access sample

frankyn:
Public Data access sample for https://cloud.google.com/storage/docs/access-public-data.

### DIFF
--- a/storage/files.rb
+++ b/storage/files.rb
@@ -133,7 +133,7 @@ def download_public_file bucket_name:, file_name:, local_path:
   file.download local_path
 
   puts "Downloaded #{file.name}"
-  # [END public_file_download]
+  # [END download_public_file]
 end
 
 def download_file_requester_pays project_id:, bucket_name:, file_name:, local_path:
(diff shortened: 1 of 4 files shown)


storage/files.rb
```diff
@@ -118,6 +118,24 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
   # [END download_file]
 end
 
+def download_public_file bucket_name:, file_name:, local_path:
+  # [START download_public_file]
+  # bucket_name = "Your Google Cloud Storage bucket name"
```
frankyn: Not necessarily the users bucket name but a allUsers readable bucket name.

storage/files.rb
```diff
@@ -118,6 +118,24 @@ def download_file project_id:, bucket_name:, file_name:, local_path:
...
+  # file_name   = "Name of a file in the Cloud Storage bucket"
+  # local_path  = "Path to local file to save"
```
frankyn: Destination path for downloaded file.

PR #229, opened 2018-01-26
Adds registry management examples

gguuss:


gguuss:
FYI -- https://console.cloud.google.com/flows/enableapi?apiid=cloudiot.googleapis.com

iot/generate_keys.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
remi: This file isn't referenced anywhere, as far as I can find

How do I use it? Could you add something to the README about this, eg. an example of running this and then calling one of the sample commands?

iot/spec/resources/README.md
```diff
@@ -0,0 +1,4 @@
+# Test public certificate files
```
remi: Thanks for adding this 😄

iot/spec/iot_spec.rb
```diff
@@ -0,0 +1,439 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "google/apis/cloudiot_v1"
+require "google/cloud/pubsub"
+require "rspec"
+require 'securerandom'
```
remi: Double Quotes everywhere `"`

*(Sorry we don't have a linter at the moment to pick these up! and some of the other syntax I'll call out)*

remi:
FYI I'm sending suggested fixes your way in a PR into your branch shortly 🤘

iot/iot.rb
```diff
@@ -0,0 +1,795 @@
...
+  parent = "projects/#{project_id}/locations/#{location_id}"
+  resource = "#{parent}/registries/#{registry_id}"
+
+  # List the devices in the provided region
+  policy = iot_client.get_registry_iam_policy(
+    resource
+  )
+
+  puts policy
+  puts policy.inspect
```
remi: Is this left-over from debugging?

I'd prefer not to `puts object` or `puts object.inspect` because it's not useful prose, the reader doesn't know what this will output

iot/iot.rb
```diff
@@ -0,0 +1,795 @@
...
+
+  # The resource name of the location associated with the project
+  parent = "projects/#{project_id}/locations/#{location_id}"
+  resource = "#{parent}/registries/#{registry_id}/devices/#{device_id}"
+
+  config = Cloudiot::DeviceConfig.new
+  config.binary_data = data
+
+  # Set configuration for the provided device
+  res = iot_client.modify_cloud_to_device_config(
```
remi: the `res` variable (result of calling `modify_cloud_to_device_config`) is unused

Is there anything useful in it to print out? If not, call the method without capturing the response in a variable

gguuss: It just has the ETAG and I was not seeing it, going ahead and removing

iot/iot.rb
gguuss: Yes, also found another where I was printing debugging.

iot/spec/iot_spec.rb
gguuss: Thanks for the suggested changes!

iot/generate_keys.sh
gguuss: Added to README

PR #232, opened 2018-02-08
Updates to IoT samples to make them more idiomatic

remi:
#229 IoT samples PR looks great! I updated it to be more idiomatic syntactically, though. TAL!

 - Lots of Rebecca 'OCD' variable/parameter assignment alignment
 - Only use `"`" double quotes
 - Updated `nil` checks to be implicit ("falseyness")
 - Updated .count checks to use `any?` enumerable predicate method
 - Updated multi-line `#each` methods to use `do` rather than `{ }` curlies
 - Rebecca's crazy aligments
 - Moved Pub/Sub topic creation helper (only used in specs) into the specs as a regular method (and moved topic cleanup into `after` block for proper teardown)
 - Updated paths to .pem files used by specs to be absolute, relative (specs weren't runnable from parent directory)
 - Did I mention I'm kind of nitpicky about alignments yet?
 - Removed all \n\n\n double newlines
 - Removed `or` keyword (should never be used, only `&&` or `||`, never `and` or `or` *[yeah, I know...]*)

### DIFF
--- a/iot/iot.rb
+++ b/iot/iot.rb
@@ -16,26 +16,6 @@
 #       method definitions in Ruby. To allow for this, code snippets in this
 #       sample are wrapped in global lambdas.
 
-# Helper for creating / deleting PubSub topics
-
-$create_pubsub_topic = -> (project_id:, topic_id:) do
-  pubsub = Google::Cloud::Pubsub.new project: project_id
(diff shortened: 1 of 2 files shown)


gguuss:
Thanks for the fixes and review! I'm going to go ahead and merge your commits.

PR #235, opened 2018-02-13
Adds list states

gguuss:
Missed one method in initial commit!

### DIFF
--- a/iot/iot.rb
+++ b/iot/iot.rb
@@ -488,8 +488,12 @@ $get_device_states = -> (project_id:, location_id:, registry_id:, device_id:) do
   res = iot_client.list_project_location_registry_device_states(
     resource
   )
-  res.device_states.each do |state|
-    puts "#{state.update_time}: #{state.binary_data}"
+  if res.device_states
+    res.device_states.each do |state|
+      puts "#{state.update_time}: #{state.binary_data}"
+    end
+  else
+    puts "No state messages"
   end
   # [END iot_get_device_states]
 end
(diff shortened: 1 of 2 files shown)


gguuss:
@frankyn - I missed an API call for listing device states. Added in this commit.

remi:
This commit only has iot.rb, it's missing the test

frankyn:
Hey @gguuss! Apologies for the delay. +1 to @remi's comment. Could you add a test for this API call?

gguuss:
Added new test for list states and included list configs as well (also missing)

gguuss:
@frankyn iot_test.rb added, I globbed in the list configs and list state into a single test because the build-up on these stupid tests is long and takes forever

frankyn:
Thanks @gguuss, the PR LGTM, could you please update the spec to hide the verbose messaging: Here's an [example](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/kms/spec/kms_spec.rb#L291-L294). 

I also see output coming from tests, please update the specs so they don't print out from your tests. 
```
Device: patches_unauth_device
	Blocked: 
	Last Event Time: 
	Last State Time: 
	Name: projects/cloud-samples-tests-ruby/locations/us-central1/registries/A872a15e7e11c56cbcreate_delete_test_patches/devices/2657738874939473
```

Apologies for the delay in review. Thanks!

frankyn:
Waiting for tests to complete to review. Thanks @gguuss!

PR #240, opened 2018-02-24
[Kokoro] Add Auth Tests

frankyn:
This is a PR to use all system tests during a Kokoro build.

### DIFF
--- a/.kokoro/system_tests.sh
+++ b/.kokoro/system_tests.sh
@@ -2,14 +2,5 @@
 
 source $KOKORO_GFILE_DIR/secrets.sh
 
-sudo apt-get update
-sudo apt-get install realpath wget
-wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64
-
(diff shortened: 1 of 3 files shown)


frankyn:
@SurferJeffAtGoogle PTAL when you have a moment.

broady:
Could you run shellcheck on this? (Should be installed on your Google workstation, but otherwise, I think this is it: https://github.com/koalaman/shellcheck)

I think we should have `set -e -u -o pipefail`, for example. (but I don't think shellcheck would pick that up)

frankyn:
@broady, I thought to not fail the script on the first failure. Do you recommend this over displaying failures instead with an overall status test?

broady:
If we expect things to fail, trap the status with something like:

```
exit_status=0
function set_failed_status {
  exit_status=1
}

./run_test || set_failed_status

exit $exit_status
```

frankyn:
@broady PTAL at the script now. Thanks!

PR #247, opened 2018-03-08
Change to endpoints managed rollout

simonz130:
We introduce a new feature where we can specify in YAML that service configuration for Endpoints is going to be managed. This means that the service will pull latest deployed service configuration automatically, rather than requiring customer to specify specific version of service configuration.

New flag:
**--rollout_strategy=managed**
Previously the only way to specify a version was using the following flag:
**--version=SERVICE_CONFIG_ID**

### DIFF
--- a/endpoints/getting-started-grpc/README.md
+++ b/endpoints/getting-started-grpc/README.md
@@ -47,7 +47,7 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
     # set your project to make commands easier
     GOOGLE_CLOUD_PROJECT=<Your Project ID>
 
-    # Print out your Config ID again, in case you missed it
+    # Print out your Service name again, in case you missed it
     gcloud endpoints services configs list --service hellogrpc.endpoints.${GOOGLE_CLOUD_PROJECT}.cloud.goog
     ```
 
--- a/endpoints/getting-started-grpc/README.md
+++ b/endpoints/getting-started-grpc/README.md
@@ -77,7 +77,6 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
     ```bash
     GOOGLE_CLOUD_PROJECT=$(curl -s "http://metadata.google.internal/computeMetadata/v1/project/project-id" -H "Metadata-Flavor: Google")
     SERVICE_NAME=hellogrpc.endpoints.${GOOGLE_CLOUD_PROJECT}.cloud.goog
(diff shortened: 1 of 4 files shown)


broady:
LGTM

endpoints/getting-started-grpc/deployment.yaml
```diff
@@ -44,7 +44,7 @@ spec:
           "-P", "9000",
           "-a", "grpc://127.0.0.1:50051",
           "-s", "SERVICE_NAME",
-          "-v", "SERVICE_CONFIG_ID",
+          "--rollout__strategy", "managed",
```
wmwong: --rollout_strategy

(Only a single underscore)

endpoints/getting-started-grpc/README.md
```diff
@@ -77,7 +77,6 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
     ```bash
```
wmwong: Up above:
- Change "gcloud service-management deploy ..." -> "gcloud endpoints services deploy ..."
- Remove the comment directly underneath about noting the configuration ID.
- Remove the command to print out the configuration ID.

endpoints/getting-started-grpc/README.md
```diff
@@ -41,14 +41,14 @@ Cloud account and [SDK](https://cloud.google.com/sdk/) configured.
...
-    gcloud service-management deploy out.pb api_config.yaml
+    gcloud endpoints services deploy out.pb api_config.yaml
     # The Config ID should be printed out, looks like: 2017-02-01r0, remember this
 
     # set your project to make commands easier
     GOOGLE_CLOUD_PROJECT=<Your Project ID>
 
     # Print out your Config ID again, in case you missed it
-    gcloud service-management configs list --service hellogrpc.endpoints.${GOOGLE_CLOUD_PROJECT}.cloud.goog
+    gcloud endpoints services configs list --service hellogrpc.endpoints.${GOOGLE_CLOUD_PROJECT}.cloud.goog
```
broady: Should you remove this config ID stuff? (and above, where it says "config ID should be printed out, looks like ...)

broady:
LGTM

endpoints/getting-started-grpc/README.md
simonz130: Config Id - yes, but this command is actually helpful to also get the service name, so I prefer to leave it - just change the documentation slightly.

PR #250, opened 2018-03-13
[Kokoro] Add CDN test

frankyn:
Add CDN test into Kokoro.

#251


frankyn:
Thanks @SurferJeffAtGoogle

Issue #251, opened 2018-03-13
Kokoro: Move ruby-docs-samples to Kokoro


frankyn:
List of tests to add

- [ ] 

- [x] - auth
- [ ] - cdn
- [ ] - datastore
- [ ] - bigquery
- [ ] - endpoints
- [ ] - iot
- [ ] - kms
- [ ] - language
- [ ] -logging
- [ ] - monitoring
- [ ] - pubsub
- [ ] - spanner
- [ ] - speech
- [ ] - storage
- [ ] - translate
- [ ] - video
- [ ] - vision
- [ ] - appengine/analytics
- [ ] - appengine/cloudsql-mysql
- [ ] - appengine/cloudsql-postgres
- [ ] - appengine/datastore
- [ ] - appengine/endpoints
- [ ] - appengine/mailgun
- [ ] - appengine/memcache
- [ ] - appengine/metadata_server
- [ ] - appengine/pubsub
- [ ] - appengine/rails-cloudsql-mysql
- [ ] - appengine/rails-cloudsql-postgresql
- [ ] - appengine/rails-hello_world
- [ ] - appengine/sendgrid
- [ ] - appengine/static_files/rails
- [ ] - appengine/static_files/sinatra
- [ ] - appengine/storage
- [ ] - appengine/twilio

frankyn:
@SurferJeffAtGoogle's team will be taking over this request. Thanks Jef!

tbpg:
All tests are enabled for Kokoro and Circle is disabled. Marking this as fixed. :tada:

frankyn:
Great work @tbpg!!!

PR #252, opened 2018-03-13
[Kokoro] Add iot sample tests

frankyn:
Add iot test into Kokoro.

#251


frankyn:
@broady merging now. Thanks for review!

PR #253, opened 2018-03-16
 Added a new Pub/Sub sample for publishing/receiving messages with custom attributes

michaelawyu:


frankyn:
@michaelawyu apologies for the delay. Could you add the new commands to the README.md?

pubsub/spec/topics_spec.rb
```diff
@@ -177,6 +177,7 @@ def expect_with_retry attempts: 5
   end
 
   it "publishes messages with batch settings" do
+    sleep(30)
```
frankyn: What if we add a `before :each` with this sleep instead?

michaelawyu: Sure :)

michaelawyu: Hi Frank! I am thinking if we can just set `config.default_sleep_interval` to `40` instead of `10`; would it be better than adding `sleep(30)` to `before :each`?

PR #254, opened 2018-03-20
Dialogflow

dizcology:
this replaces https://github.com/GoogleCloudPlatform/ruby-docs-samples/pull/239


frankyn:
Acknowledging request and passing review to @remi ML related PRs. Thanks @dizcology and @remi!

remi:
FYI/Update: I'm trying these samples out locally today (to port them to another lang). I'll add new comments today if I stumble upon anything

dialogflow/detect_intent_texts.rb
```diff
@@ -0,0 +1,65 @@
...
+end
+
+
+if __FILE__ == $PROGRAM_NAME
+  project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+  texts = ARGV
+
+  if !texts.empty?
+    detect_intent_texts project_id: project_id,
+                        session_id: session_id,
```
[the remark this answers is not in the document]
remi: Nvm talked to @sirtorry and got the rundown on session IDs 😄 

^ this is still a bug though

dialogflow/entity_type_management.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  entity_types_client = Google::Cloud::Dialogflow::EntityTypes.new
+  entity_type_path = entity_types_client.class.entity_type_path project_id, entity_type_id
+
+  response = entity_types_client.delete_entity_type entity_type_path
+  # [END dialogflow_delete_entity_type]
+end
+
+
+# Helper
```
beccca: You don't need a comment for these helper methods

When reading a spec, my expectation is already that I might see helpers ~ spec files are often laid out like so:

```
describe
  def helper1
  def heper2
  before
  after
  example1
  example2
  ...
```

dialogflow/session_entity_type_management.rb
```diff
@@ -0,0 +1,123 @@
...
+  session_entity_type = { 
+    name:                 session_entity_type_name,
+    entities:             entities,
+    # the modes are either ENTITY_OVERRIDE_MODE_OVERRIDE or ENTITY_OVERRIDE_MODE_SUPPLEMENT
+    entity_override_mode: :ENTITY_OVERRIDE_MODE_OVERRIDE
+  }
+
+  response = session_entity_types_client.create_session_entity_type session_path, session_entity_type
+
+  puts "SessionEntityType created: #{response}"
```
beccca: What is the result of outputing the `response` object?

Vote: don't print out `response`

dialogflow/entity_management.rb
```diff
@@ -0,0 +1,104 @@
...
+  require "google/cloud/dialogflow"
+
+  entity_types_client = Google::Cloud::Dialogflow::EntityTypes.new
+  entity_type_path = entity_types_client.class.entity_type_path project_id, entity_type_id
+
+  entity = { value: entity_value, synonyms: synonyms }
+
+  response = entity_types_client.batch_create_entities entity_type_path, [entity]
+
+  puts "Entity created: #{response}"
```
beccca: What does this print out?

As this is a batch rpc method, I'm guessing that it returns multiple response objects, 1 per entity?

*Ah! Nope!*

`BatchCreateEntities` returns an Operation, it's a LRO!

https://cloud.google.com/dialogflow-enterprise/docs/reference/rpc/google.cloud.dialogflow.v2beta1#google.cloud.dialogflow.v2beta1.EntityTypes.BatchCreateEntities-table

So this sample should probably follow the usual template for LRO samples (?):
 - print out operation info
 - wait for operation to complete
 - print that it completed

beccca: The op returns Empty, so there's no .result field on it to print out, but the existing code only *kicks off the LRO*, there's no knowing that it completed

dialogflow/detect_intent_audio.rb
```diff
@@ -0,0 +1,78 @@
...
+    sample_rate_hertz: 16000,
+    language_code: language_code
+  }
+
+  query_input = { audio_config: audio_config }
+
+  response = session_client.detect_intent session, query_input, input_audio: input_audio
+  query_result = response.query_result
+
+  puts "=" * 20
```
beccca: Let's not use this cute, fancy Ruby feature to add separators ;)

It took me a moment to understand what this is doing, at first glance it's confusing. The sample is complex enough, let's not do this :)

We're only printing out 4 lines per Intent, so I wouldn't add a separator at all for this one

I tend to stay away from anything in the sample code that exists only to format the CLI output, eg. I also prefer to not see `\t` and `\n` because it's irrelevant to the code & just adds extra syntax fluff

beccca: For this one, in particular, `"=" * 20` is too _**cleaver**_ :P

dialogflow/detect_intent_texts.rb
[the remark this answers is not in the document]
dizcology: done.

dialogflow/entity_type_management.rb
```diff
@@ -0,0 +1,116 @@
...
+  # [START dialogflow_delete_entity_type]
+  # project_id = "Your Google Cloud project ID"
+  # entity_type_id = "Existing Entity Type ID"
+  
+  require "google/cloud/dialogflow"
+
+  entity_types_client = Google::Cloud::Dialogflow::EntityTypes.new
+  entity_type_path = entity_types_client.class.entity_type_path project_id, entity_type_id
+
+  response = entity_types_client.delete_entity_type entity_type_path
```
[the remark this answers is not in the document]
dizcology: done.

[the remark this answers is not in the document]
dizcology: done.

dialogflow/session_entity_type_management.rb
dizcology: Agreed - for Ruby this prints something that's not useful.

dialogflow/entity_type_management.rb
```diff
@@ -0,0 +1,116 @@
...
+  # [END dialogflow_delete_entity_type]
+end
+
+
+# Helper
+def get_entity_type_ids project_id:, display_name:
```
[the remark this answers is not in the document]
dizcology: I am sort of on the fence on this one.  This appears to be used only as a helper for the tests, but in fact is a useful helper for the developers as well.  

This fills in the gap that entity types are sometimes identified with their display names (e.g. when creating session entity types) and some other times with their ids (e.g. when deleting an entity type).

That being said, the chance of developers discovering this helper method seems low and I'll just move it into the spec.

dialogflow/detect_intent_audio.rb
dizcology: done.

dialogflow/entity_management.rb
dizcology: done.

the operation has at least the `done?` method.

dialogflow/entity_type_management.rb
dizcology: done.  moved the helpers into spec/spec_helper.rb

dialogflow/Gemfile
```diff
@@ -0,0 +1,22 @@
+# Copyright 2016 Google, Inc
```
beccca: 2018 (here & wherever else)

dizcology: done

dialogflow/detect_intent_texts.rb
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+
+require "securerandom"
+
+def detect_intent_texts project_id:, session_id:, texts:, language_code:
+  # [START dialogflow_detect_intent_texts]
+  # project_id = "Your Google Cloud project ID"
+  # session_id = "mysession"
+  # language_code = "en-US"
```
beccca: Missing `texts` variable. So I can't see what kind of input is expected.

dialogflow/detect_intent_audio.rb
```diff
@@ -0,0 +1,77 @@
...
+
+
+require "securerandom"
+
+def detect_intent_audio project_id:, session_id:, audio_file_path:,
+                        language_code:
+  # [START dialogflow_detect_intent_audio]
+  # project_id = "Your Google Cloud project ID"
+  # session_id = "mysession"
+  # language_code = "en-US"
```
beccca: Missing audio_file_path parameter

dialogflow/detect_intent_stream.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+  end
+end
+
+def detect_intent_stream project_id:, session_id:, audio_file_path:,
+                        language_code:
+  # [START dialogflow_detect_intent_stream]
+  # project_id = "Your Google Cloud project ID"
+  # session_id = "mysession"
+  # language_code = "en-US"
```
beccca: Missing audio_file_path parameter

dialogflow/spec/spec_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+  entity_types_client = Google::Cloud::Dialogflow::EntityTypes.new
+  parent = entity_types_client.class.project_agent_path project_id
+
+  entity_types = entity_types_client.list_entity_types parent
+
+  selected_entity_types = entity_types.select { |entity_type| entity_type.display_name == display_name }
+
+  entity_type_ids = selected_entity_types.map { |entity_type| entity_type.name.split("/").last }
+
+  return entity_type_ids
```
beccca: You should never use `return` in Ruby

(As a general rule. Unless short-circuiting.)

dialogflow/detect_intent_texts.rb
beccca: Also, align =

dialogflow/detect_intent_texts.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+  # [END dialogflow_detect_intent_texts]
+end
+
+
+if __FILE__ == $PROGRAM_NAME
+  project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+  texts = ARGV
+
+  if !texts.empty?
```
beccca: `texts.any?`

dialogflow/entity_management.rb
```diff
@@ -0,0 +1,111 @@
...
+
+  entity_types_client = Google::Cloud::Dialogflow::EntityTypes.new
+  entity_type_path = entity_types_client.class.entity_type_path project_id, entity_type_id
+
+  entity = { value: entity_value, synonyms: synonyms }
+
+  operation = entity_types_client.batch_create_entities entity_type_path, [entity]
+
+  puts "Waiting for the entity creation operation to complete."
+  while !operation.done?
```
beccca: `operation.wait_until_done!`

dialogflow/intent_management.rb
```diff
@@ -0,0 +1,122 @@
...
+  # display_name = "New Display Name"
+  # message_text = "some message text"
+  # training_phrases_parts = ["part1", "part2"]
+  
+  require "google/cloud/dialogflow"
+
+  intents_client = Google::Cloud::Dialogflow::Intents.new
+  parent = intents_client.class.project_agent_path project_id
+
+  intent = {
```
beccca:
```ruby
intent = { 
  display_name: display_name,
  messages: [{ text: { text: [message_text] } }],
  training_phrases: training_phrases_parts.map { |part|
    { parts: [{ text: part }]}
  } # <--- drop this curly onto a newline ~~ "appers" as if it's broken/unclosed, as-is
}
```

dialogflow/session_entity_type_management.rb
```diff
@@ -0,0 +1,123 @@
...
+  session_entity_type = { 
+    name:                 session_entity_type_name,
+    entities:             entities,
+    # the modes are either ENTITY_OVERRIDE_MODE_OVERRIDE or ENTITY_OVERRIDE_MODE_SUPPLEMENT
+    entity_override_mode: :ENTITY_OVERRIDE_MODE_OVERRIDE
+  }
+
+  response = session_entity_types_client.create_session_entity_type session_path, session_entity_type
+
+  puts "SessionEntityType name:    #{response.name}"
```
beccca: remove extra whitespace in output

dialogflow/spec/context_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "rspec"
+require "google/cloud/dialogflow"
+
+require_relative "../context_management"
+require_relative "../detect_intent_texts"
```
beccca: Why does the Context Management spec require the Detect Intents Texts sample? Typo? Leftover from old code?

beccca: oh, I see you're using it to "create a session" ...

dialogflow/spec/context_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+describe "Context Management" do
+
+  before do
+    @project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+    @session_id = "fake_session_for_testing"
+    @context_id = "fake_context_for_testing"
+  end
+
+  example "create context" do
+    # create a session
```
beccca: What happens if you `create_context` without calling this first?

For me, my Context Management create test does this:

 1. var session = "randomwhatever"
 1. verify LIST does NOT include "somecontextid"
 1. create "somecontextid"
 1. verify LIST does now include "somecontextid"

Session is arbitrary, it can be anything 😄

beccca: Vote: don't call the detect intent sample from this or other 'management' specs

dialogflow/spec/context_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      /#{@session_id}.*#{@context_id}/m
+    ).to_stdout
+  end
+
+  example "delete context" do
+    delete_context project_id: @project_id,
+                   session_id: @session_id,
+                   context_id: @context_id
+
+    expect {
```
beccca: How do we know that `list_contexts` ever output @context_id in the first place?

```
verify ! exist

create thing

verify exist
```

beccca: Example of a test which follows this pattern to verify behavior of a CREATE action
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/storage/spec/files_spec.rb#L147-L162

or DELETE action
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/storage/spec/buckets_spec.rb#L216-L225

dialogflow/spec/create_delete_entity_spec.rb
```diff
@@ -0,0 +1,102 @@
+# Copyright 2018 Google, Inc
```
beccca: The samples being tested are `entity_management.rb` so I expect this to be called `spec/entity_management_spec.rb`

dialogflow/spec/create_delete_entity_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    delete_entity project_id: @project_id, entity_type_id: @entity_type_id,
+                  entity_value: @entity_value_1
+    delete_entity project_id: @project_id, entity_type_id: @entity_type_id,
+                  entity_value: @entity_value_2
+
+    expect {
+      list_entities project_id: @project_id, entity_type_id: @entity_type_id
+    }.to output(
+      ""
```
beccca: Rather than asserting that the output HAS an empty string, how about asserting that the output DOESN'T HAVE a string that you're looking for (eg. a string that was there when you called list_entities before doing deletion)

Asserting that an object does NOT appear in a list gets across the intent of the assertion, as you expect the item to no longer be there (rather than expecting an "" to be present)

dialogflow/spec/detect_intent_texts_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  example "detect intent from texts" do
+    expect {
+      detect_intent_texts project_id: @project_id,
+                          session_id: @session_id,
+                          texts: @texts,
+                          language_code: @language_code
+    }.to output(
+      /All set!/
```
beccca: Nice! The 3 detect intent specs a are short and sweet! dig it 👍

dialogflow/spec/context_spec.rb
beccca: Note: when I scrolled down, I saw that you use this pattern in most of your Create/Delete tests, but not this Context spec :)

dialogflow/spec/context_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    # create a session
+    detect_intent_texts project_id:    @project_id,
+                        session_id:    @session_id,
+                        texts:         ["hi"],
+                        language_code: "en-US" 
+    expect {
+      create_context project_id: @project_id,
+                     session_id: @session_id,
+                     context_id: @context_id
+    }.to output(
```
beccca: This tests that create_context outputs the right STDOUT but doesn't necessarily prove that the context exists (which checking List output is very helpful for)

dialogflow/spec/spec_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def get_entity_type_ids project_id:, display_name:
+  # project_id = "Your Google Cloud project ID"
+  # display_name = "Existing Display Name"
+  
+  require "google/cloud/dialogflow"
```
beccca: Move the `require` to the top of the file.

`require` should never be inside of a class/method/module (unless you have a very good reason)

^--- we put them into our snippet methods so the require will be inside of region tag blocks

Also, please remove the placeholder variable comments, these are part of the snippet code rubric and not idiomatic Ruby

```ruby
# LICENSE

require "google/cloud/dialogflow"

def get_entity_type_ids project_id, display_name
  client       = Google::Cloud::Dialogflow::EntityTypes.new
  parent       = client.class.project_agent_path project_id
  entity_types = entity_types_client.list_entity_types parent

  entity_types.select { |et| et.display_name == display_name }.
                  map { |et| et.name.split("/").last }
end

# ...
```

dialogflow/spec/spec_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def get_entity_type_ids project_id:, display_name:
+  # project_id = "Your Google Cloud project ID"
```
beccca: Remove these comments

dialogflow/spec/spec_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  selected_entity_types = entity_types.select { |entity_type| entity_type.display_name == display_name }
+
+  entity_type_ids = selected_entity_types.map { |entity_type| entity_type.name.split("/").last }
+
+  return entity_type_ids
+end
+
+def get_intent_ids project_id:, display_name:
+  # project_id = "Your Google Cloud project ID"
```
beccca: Remove these comments

dialogflow/detect_intent_texts.rb
dizcology: done except for = alignment.

dialogflow/detect_intent_audio.rb
dizcology: done

dialogflow/detect_intent_stream.rb
dizcology: done

dialogflow/spec/spec_helper.rb
dizcology: done

dialogflow/spec/spec_helper.rb
dizcology: done.

dialogflow/spec/spec_helper.rb
dizcology: done

dialogflow/spec/spec_helper.rb
dizcology: done

dialogflow/detect_intent_texts.rb
dizcology: done

dialogflow/entity_management.rb
dizcology: done

dialogflow/intent_management.rb
dizcology: done

dialogflow/session_entity_type_management.rb
dizcology: done

dialogflow/spec/context_spec.rb
dizcology: done

dialogflow/spec/create_delete_entity_spec.rb
dizcology: done

dialogflow/spec/context_spec.rb
dizcology: done

dizcology:
review comments addressed except:

1. variable assignment alignment
2. streaming sample (which indeed was mostly copied from the analogous sample for Speech API)

dialogflow/detect_intent_stream.rb
```diff
@@ -0,0 +1,132 @@
...
+  request_queue  = EnumeratorQueue.new(session_client)
+
+  # The first request needs to be the configuration.
+  request_queue.push(streaming_config)
+
+  # Consume the queue and process responses in a separate thread
+  Thread.new do
+    session_client.streaming_detect_intent(request_queue.each_item).each do |response|
+      if response.recognition_result
+        puts "Intermediate transcript: #{response.recognition_result.transcript}\n"
```
frankyn: out of curiosity, does intermediate transcript print output right now?

dialogflow/detect_intent_stream.rb
```diff
@@ -0,0 +1,132 @@
...
+    audio_file = File.open(audio_file_path, "rb")
+      while true
+        chunk = audio_file.read 4096
+        break if not chunk
+        request_queue.push({ input_audio: chunk})
+        sleep 0.5
+      end
+  ensure
+    audio_file.close
+    request_queue.push(session_client)
```
frankyn: I'd add a comment stating that pushing the `session_client` ends the streaming queue.

dialogflow/spec/context_management_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    ).to_stdout
+    expect {
+      list_contexts project_id: @project_id, session_id: @session_id
+    }.to output(
+      /#{@context_id}/
+    ).to_stdout
+  end
+
+  example "delete context" do
+    delete_context project_id: @project_id,
```
beccca: **tl;dr** *These tests run in random order so `example "create context"` may not have run before this, so there is no context to delete. Create a context right above this and you're good to go!* 👍 

----
`<rant>`

There's no Context to delete here.

Seeing the `delete`, I naturally looked for:
 1. Create context in `example()`
 1. Create context in `before()`

But found none 😥 

Ruby tests are run in random order _(this has become the default for the 2 popular testing frameworks, iirc)_ so you can't guarantee this will have anything to delete. The above `example "create context"` may not have been run yet.

Because of this, it's natural for Rubyists to create prerequisite objects either:
 - (a) in the `example` _(aka `it`)_ block . 
 - (b) in `before` so it runs before every test for scenarios where every test is going to operate on a new instance of an object. There is also an `after`.
 - (c) in `before :all do ... end` which runs *once* before all of the `example`s are run, eg. to create some objects multiple tests require and the tests promise not to mutate these objects. There is also an `after :all`.

`</rant>`

dialogflow/detect_intent_stream.rb
dizcology: it does:
```
% ruby detect_intent_stream.rb resources/book_a_room.wav
Session path: projects/rising-sea-112358/agent/sessions/80f3f76f-3146-48a4-9554-a10727e6da7f
Intermediate transcript: book
Intermediate transcript: buca
Intermediate transcript: book of
Intermediate transcript: book a room
Intermediate transcript: book a room
Query text:        book a room
Intent detected:   room.reservation
Intent confidence: 1.0
Fulfillment text:  I can help with that. Where would you like to reserve a room?
```

dialogflow/detect_intent_stream.rb
dizcology: done.

dialogflow/spec/context_management_spec.rb
dizcology: done.  

thanks for pointing out that the tests run in random order (why is that?)

PR #256, opened 2018-03-22
BigQuery Data Transfer: Add quickstart.

tswast:
Quickstart lists available data sources.

To be published in the docs at https://cloud.google.com/bigquery/transfer/docs/reference/libraries

### DIFF
--- a/bigquerydatatransfer/quickstart.rb
+++ b/bigquerydatatransfer/quickstart.rb
@@ -16,7 +16,8 @@
 # Imports the Google Cloud client library
 require "google/cloud/bigquery/data_transfer"
 
-project_id = "YOUR_PROJECT_ID"  # TODO: Update to your project ID.
+# Your Google Cloud Platform project ID
+project_id = "YOUR_PROJECT_ID"
 
 # Instantiate a client
 BigQueryDataTransfer = Google::Cloud::Bigquery::DataTransfer # Alias the module
@@ -24,16 +25,17 @@ data_transfer = BigQueryDataTransfer.new
 
 # Get the full path to your project.
 project_path = BigQueryDataTransfer::V1::DataTransferServiceClient.project_path(
-  project_id)
+  project_id
+)
 
 puts "Supported Data Sources:"
 
 # Iterate over all possible data sources.
 data_transfer.list_data_sources(project_path).each do |data_source|
-  puts "#{data_source.display_name}:"
-  puts "\tID: #{data_source.data_source_id}"
-  puts "\tFull path: #{data_source.name}"
-  puts "\tDescription: #{data_source.description}"
+  puts "Data source: #{data_source.display_name}"
+  puts "ID: #{data_source.data_source_id}"
+  puts "Full path: #{data_source.name}"
+  puts "Description: #{data_source.description}"
 end
 # [END bigquerydatatransfer_quickstart]
 
--- a/bigquerydatatransfer/spec/quickstart_spec.rb
+++ b/bigquerydatatransfer/spec/quickstart_spec.rb
@@ -30,7 +30,7 @@ describe "BigQuery Data Transfer Quickstart" do
     expect {
       load File.expand_path("../quickstart.rb", __dir__)
     }.to output(
-      /Supported Data Sources:\n/
+      /ID: youtube_channel\n/
     ).to_stdout
    end
 


tswast:
Re: Circle failure. Do we still support 2.2.8? I get a nasty warning from the client libraries when I use that version.

frankyn:
Sharing an internal doc.

tswast:
Okay, rebased. Hopefully Circle should be happy now.

tswast:
@frankyn Could you enable the DTS API on this project?

```
BigQuery Data Transfer Quickstart
  lists data sources (FAILED - 1)

Failures:

  1) BigQuery Data Transfer Quickstart lists data sources
     Failure/Error: load File.expand_path("../quickstart.rb", __dir__)
     
     Google::Gax::RetryError:
       GaxError Exception occurred in retry method that was not classified as transient, caused by 7:BigQuery Data Transfer API has not been used in project cloud-samples-tests-ruby before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/bigquerydatatransfer.googleapis.com/overview?project=cloud-samples-tests-ruby then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```

Aside: We can enable the API ourselves, which is enough for this sample to work, but we'll also have to "[enroll](https://cloud.google.com/bigquery/docs/enable-transfer-service)" once we have some samples that actually need to kick off a transfer.

tswast:
@frankyn Nevermind, I see I have permissions on the project now.

tswast:
@alixhami or @remi Could you review this BQ-DTS Ruby sample?

tswast:
Ping @remi. Please review.

bigquerydatatransfer/quickstart.rb
```diff
@@ -0,0 +1,39 @@
...
+require "google/cloud/bigquery/data_transfer"
+
+project_id = "YOUR_PROJECT_ID"  # TODO: Update to your project ID.
+
+# Instantiate a client
+BigQueryDataTransfer = Google::Cloud::Bigquery::DataTransfer # Alias the module
+data_transfer = BigQueryDataTransfer.new
+
+# Get the full path to your project.
+project_path = BigQueryDataTransfer::V1::DataTransferServiceClient.project_path(
```
beccca: Uggghhhhh.... this literally just returns `"projects/#{project_id}"`

I HATE these project path helpers. Disgusting.

LGTM though ~ we're stuck with this for now

bigquerydatatransfer/quickstart.rb
```diff
@@ -0,0 +1,39 @@
...
+  project_id)
+
+puts "Supported Data Sources:"
+
+# Iterate over all possible data sources.
+data_transfer.list_data_sources(project_path).each do |data_source|
+  puts "#{data_source.display_name}:"
```
beccca: The other ones all follow a `Value Description: #{how to get value}` pattern, let's use that for the display name too

I vote the following to be the most readable / easiest to consume _(basically I just follow the same pattern for all values, actual output STDOUT is a low priority)_ 

```rb
data_transfer.list_data_sources(project_path).each do |data_source|
  puts "Data source: #{data_source.display_name}"
  puts "ID: #{data_source.data_source_id}"
  puts "Full path: #{data_source.name}"
  puts "Description: #{data_source.description}"
end
```

bigquerydatatransfer/quickstart.rb
```diff
@@ -0,0 +1,39 @@
...
+
+project_id = "YOUR_PROJECT_ID"  # TODO: Update to your project ID.
+
+# Instantiate a client
+BigQueryDataTransfer = Google::Cloud::Bigquery::DataTransfer # Alias the module
+data_transfer = BigQueryDataTransfer.new
+
+# Get the full path to your project.
+project_path = BigQueryDataTransfer::V1::DataTransferServiceClient.project_path(
+  project_id)
```
beccca: Push parens to own line

```rb
project_path = BigQueryDataTransfer::V1::DataTransferServiceClient.project_path(
  project_id
)
```

bigquerydatatransfer/spec/quickstart_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    expect(Google::Cloud::Bigquery::DataTransfer::V1::DataTransferServiceClient).to(
+      receive(:project_path).
+      with("YOUR_PROJECT_ID").
+      and_return(project_path))
+
+    expect {
+      load File.expand_path("../quickstart.rb", __dir__)
+    }.to output(
+      /Supported Data Sources:\n/
```
beccca: Please test a data value from a successful API response

Issue #246, opened 2018-03-05
BigQuery: Sample is included in the wrong section of the docs


alixhami:
The [sample](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/bigquery/tables.rb) for creating a table shown [here](https://cloud.google.com/bigquery/docs/tables#bigquery-create-table-ruby) is the "Creating an empty table with a schema definition" section, but the sample does not have a schema specified.

To fix this:
+ Move this sample to the section above: "Creating an empty table without a schema definition"
+ Create a new sample with the schema specified for the "Creating an empty table with a schema definition" section

PR #269, opened 2018-04-03
BigQuery: Adds sample for creating a table with a schema

alixhami:
Resolves https://github.com/GoogleCloudPlatform/ruby-docs-samples/issues/246

### DIFF
--- a/bigquery/tables.rb
+++ b/bigquery/tables.rb
@@ -12,8 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-def create_table project_id:, dataset_id:, table_id:
+def create_table_with_schema project_id:, dataset_id:, table_id:
   # [START create_table]
+  # project_id = "Your Google Cloud project ID"
+  # dataset_id = "ID of the dataset to create table in"
+  # table_id   = "ID of the table to create"
+
+  require "google/cloud/bigquery"
+
+  bigquery = Google::Cloud::Bigquery.new project: project_id
+  dataset  = bigquery.dataset dataset_id
+
+  table    = dataset.create_table table_id do |schema|
+    schema.string  "full_name", mode: :required
+    schema.integer "age",       mode: :required
+  end
+
+  puts "Created table: #{table_id}"
+  # [END create_table]
+end
+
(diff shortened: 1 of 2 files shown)


frankyn:
Passing PR review for BD/ML to @remi.

bigquery/tables.rb
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-def create_table project_id:, dataset_id:, table_id:
+def create_table_with_schema project_id:, dataset_id:, table_id:
   # [START create_table]
```
tswast: Per our conversation yesterday, you're going to update the region tags in a separate PR, right?

bigquery/tables.rb
```diff
@@ -23,12 +23,32 @@ def create_table project_id:, dataset_id:, table_id:
   bigquery = Google::Cloud::Bigquery.new project: project_id
   dataset  = bigquery.dataset dataset_id
 
-  dataset.create_table table_id
+  table    = dataset.create_table table_id do |schema|
```
tswast: Maybe change `|schema|` to `|updater|` or `|table|`?

The block for `create_table` is actually a "Table::Updater" (I see that's not documented in the API docs, so I just created https://github.com/GoogleCloudPlatform/google-cloud-ruby/pull/2038). It lets you set any property on the table, so more than just add schema fields. \Ffor example, it'll be how you set the encryption key once that feature is released).

bigquery/tables.rb
```diff
@@ -23,12 +23,32 @@ def create_table project_id:, dataset_id:, table_id:
...
+    schema.string  "full_name", mode: :required
+    schema.integer "age",       mode: :required
+  end
 
   puts "Created table: #{table_id}"
   # [END create_table]
 end
 
+def create_table_without_schema project_id:, dataset_id:, table_id:
+  # [START create_table_without_schema]
```
tswast: Missing `bigquery_` in this region tag (I see it in END but not at START)

bigquery/tables.rb
alixhami: Yeah it currently doesn't have a branch specified, so changing the region tag now would break the link

PR #290, opened 2018-06-12
Fix Firestore region tags

ryanmats:


frankyn:
Merge when you're ready.

PR #295, opened 2018-06-20
kokoro: de-flake some dialogflow tests and enable logging tests

tbpg:
Part of #251.


frankyn:
Merging as kokoro passed and test logs are now clean for dialog flow.

PR #297, opened 2018-06-25
kokoro: always exit 0 when setting CHANGED_DIRS

tbpg:
Part of #251.


tbpg:
PR tests pass. But, this is for periodic system tests with no changes. Will keep an eye on it.

PR #302, opened 2018-07-02
[Text-to-Speech] Code samples

alixhami:
Code samples based on [Python](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/texttospeech/cloud-client). Gem is not released because the API is not yet available, so these were tested using a local version of the gem. I will add a Gemfile once the gem is released.


alixhami:
The gem is now released so I updated to use the public gem.

texttospeech/synthesize_file.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  puts "Audio content written to file 'output.mp3'"
+  # [END tts_synthesize_ssml_file]
+end
+
+if __FILE__ == $PROGRAM_NAME
+  command   = ARGV.shift
+  file_path = ARGV.shift
+
+  if command == "--text"
```
frankyn: Update command to follow other samples:

For example, instead of `--text` use `text`.

tbpg:
I enabled the API and tests are passing. :+1:

beccca:
👍

PR #307, opened 2018-07-20
kokoro: enable remaining tests

tbpg:
This PR finalizes running all tests in Kokoro. There are 3 E2E tests:
* `appengine/cloudsql-mysql`
* `appengine/cloudsql-postgres`
* `appengine/metadata_server`

These test provide sufficient coverage to make sure AppEngine works without having to deploy >10 AppEngine Flexible applications (would easily take over an hour).

Part of #251. Next step: disable Circle.


tbpg:
Kokoro failure was caused by quota. The new E2E tests passed (Ruby 2.5).

PR #315, opened 2018-07-26
dialogflow: make tests more robust

tbpg:
spec/session_entity_type_management_spec.rb fails sometimes if the
delete test is run without the create test (for example, when quota
issues cause the test to fail).

Part of #251.


tbpg:
Kokoro ran into quota issues, retried, and finished successfully (unlike before this PR). So, this appears to work.

PR #320, opened 2018-07-28
[Vision] Add missing project argument

alixhami:
Project was being provided but not used, so I added it to the client initialization.

### DIFF
--- a/vision/detect_pdf.rb
+++ b/vision/detect_pdf.rb
@@ -21,7 +21,7 @@ def detect_pdf_gcs gcs_source_uri:, gcs_destination_uri:, project_id:
   require "google/cloud/vision"
   require "google/cloud/storage"
 
-  vision = Google::Cloud::Vision.new
+  vision = Google::Cloud::Vision.new project: project_id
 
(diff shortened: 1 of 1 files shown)


vision/detect_pdf.rb
```diff
@@ -21,7 +21,7 @@ def detect_pdf_gcs gcs_source_uri:, gcs_destination_uri:, project_id:
   require "google/cloud/vision"
   require "google/cloud/storage"
 
-  vision = Google::Cloud::Vision.new
+  vision = Google::Cloud::Vision.new project: project_id
```
frankyn: IIRC, a project id is unnecessary as the Vision API doesn't support it in a request. Project id was accidentally added to the client library.

However, the project could be added to `Google::Cloud::Storage.new`.

alixhami: Should I remove the project as an argument all together? I was just basing the arguments off another sample

frankyn: Storage could still use it so I'd say leave it.

PR #329, opened 2018-08-06
Remove unused BigQuery snippet

alixhami:


alixhami:
Thanks @frankyn! It is not being used on cloudsite and there are no plans to include it in the future, so I am removing it across languages.

PR #332, opened 2018-08-13
update region tags

averikitsch:
Updating and removing region tags only

### DIFF
--- a/appengine/cloudsql-postgres/app.yaml
+++ b/appengine/cloudsql-postgres/app.yaml
@@ -17,7 +17,7 @@ env: flex
 entrypoint: bundle exec ruby app.rb -p $PORT
 
 # [START gae_flex_postgres_env]
-gae_flex_postgres_env:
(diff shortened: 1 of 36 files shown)


appengine/cloudsql-postgres/app.yaml
```diff
@@ -16,15 +16,15 @@ runtime: ruby
 env: flex
 entrypoint: bundle exec ruby app.rb -p $PORT
 
-# [START env_variables]
-env_variables:
```
tbpg: Need to revert this line.

appengine/sendgrid/app.yaml
```diff
@@ -16,8 +16,8 @@ runtime: ruby
 env: flex
 entrypoint: bundle exec ruby app.rb
 
-# [START env_variables]
-env_variables:
```
tbpg: Need to revert this line.

PR #345, opened 2018-09-11
New BigQuery samples for loading Orc from GCS

alixhami:
+ Adds new BigQuery samples for loading Orc data from GCS
+ Fixes the tests using the us-states GCS data to not use the `@table` fixture, because it has a schema defined that conflicts with the us-states data


frankyn:
Summoning @GoogleCloudPlatform/yoshi-ruby for this review.

PR #348, opened 2018-09-18
appengine/datastore: use datastore lib

### LINKED ISSUE #347: Undefined local variable in Appengine datastore app
Hello, 

I am a newbie on the Google Cloud platform.
I'm running an example of Google Cloud Platform and I'm having an issue registering an issue.

The following code is not executed:
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/appengine/datastore/app.rb#L22

The variables in dataset are not initialized.
So when I run the application, the following error occurs.

undefined local variable or method `dataset '

My development environment is ruby ​​2.4 and MacOS.

If you report wrong, please let me know.

tbpg:
Fixes #347.


appengine/datastore/app.rb
```diff
@@ -18,14 +18,15 @@
 require "google/cloud/datastore"
 
 get "/" do
+  datastore = Google::Cloud::Datastore.new
+
   # Save visit in Datastore
-  visit = dataset.entity "Visit"
+  visit = datastore.entity "Visit"
```
frankyn: Out of scope and okay to decline but it would make the sample cleaner.
[Example](http://googlecloudplatform.github.io/google-cloud-ruby/docs/google-cloud-datastore/v1.4.3/Google/Cloud/Datastore/Dataset.html#save-instance_method) shows how to use a Ruby block which gives context to the entity visit when setting `user_ip` and `timestamp`.

tbpg: Done.

tbpg:
Thanks!

PR #349, opened 2018-09-19
Adds speech samples

alixhami:
New speech samples for
+ Model selection
+ Enhanced phone model
+ Auto punctuation


speech/speech_samples.rb
```diff
@@ -250,6 +250,117 @@ def speech_streaming_recognize audio_file_path: nil
 # [END speech_transcribe_streaming]
 end
 
+def speech_transcribe_auto_punctuation audio_file_path: nil
+  # [START speech_transcribe_auto_punctuation]
+  # audio_file_path = "path/to/audio.wav"
```
beccasaurus: nit: This one is missing `\n` between placeholder variable & import

alixhami: Oops missed that. Fixed.

PR #355, opened 2018-10-09
Add Cloud Spanner DML/PDML samples.

jsimonweb:


frankyn:
@jsimonweb are these the same samples I approved internally?

PR #367, opened 2018-11-15
add video API speech transcription sample;

dizcology:
modeled after: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/video/cloud-client/analyze/beta_snippets.py#L32


frankyn:
Should have done this earlier. Deferring my review as @beccasaurus is the ML reviewer for Ruby.

PR #369, opened 2018-11-20
Iot commands

gguuss:
Adds commands example / test to Cloud IoT Sample.


iot/iot.rb
```diff
@@ -626,6 +626,38 @@
...
+  iot_client = Cloudiot::CloudIotService.new
+  iot_client.authorization = Google::Auth.get_application_default(
+    "https://www.googleapis.com/auth/cloud-platform"
+  )
+
+  # The resource name of the location associated with the project
+  parent   = "projects/#{project_id}/locations/#{location_id}"
+  resource = "#{parent}/registries/#{registry_id}/devices/#{device_id}"
+
+  commandReq = Cloudiot::SendCommandToDeviceRequest.new
```
frankyn: s/commandReq/command_request/

iot/iot.rb
```diff
@@ -626,6 +626,38 @@
...
+  commandReq.binary_data = data
+
+  # Set configuration for the provided device
+  res = iot_client.send_project_location_registry_group_device_command_to_device resource, commandReq
```
frankyn: s/res/response/g

gguuss: The response is actually unused, removing completely, renaming elsewhere.

iot/iot.rb
gguuss: Done.

PR #371, opened 2018-11-30
update vision version

TheRoyalTnetennba:


quartzmo:
I would recommend excluding `Gemfile.lock` in `.gitignore` [as we do](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-vision/.gitignore#L1) in the google-cloud-ruby packages.

blowmage:
I can see excluding the Gemfile.lock file if a narrow version was set in Gemfile. Right now it is unbound.

PR #375, opened 2018-12-05
Update detect_safe_search.rb

nnegrey:


nnegrey:
Oh, I'm not a ruby person. I just figured I could easily add a missing line from our other samples.

PR #380, opened 2018-12-21
Convert Cloud KMS Samples to GAPIC client library

bdhess:


bdhess:
cc @DanSanche

bdhess:
I'm not sure how quickstart substitution is supposed to work in Ruby, but that's the only thing that appears to be erroring:

```
  1) Key Management Service Quickstart can list global key rings by name
     Failure/Error: load File.expand_path("../quickstart.rb", __dir__)

     Google::Gax::RetryError:
       GaxError Exception occurred in retry method that was not classified as transient, caused by 5:Project 'YOUR_PROJECT_ID' not found.
```

bshaffer:
The tests aren't replacing `YOUR_PROJECT_ID` in the quickstart tests

bshaffer:
Try folllowing [these other quickstart examples](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/d16579d09f0cd58ccdd47277c385f8abcc513a0d/spanner/spec/quickstart_spec.rb#L42) and use `expect` to return the instance of the KMS client that uses the right Project ID, e.g.:

```rb
@kms = CloudKMS::KeyManagementServiceClient.new project: test_project_id
expect(CloudKMS::KeyManagementServiceClient).to receive(:new).
                                  with(project: "YOUR_PROJECT_ID").
                                  and_return(@kms)
```

PR #383, opened 2019-01-02
update product_search to use new instance helpers and fix spanner tests

TheRoyalTnetennba:
Update ProductSearch samples to use new instance helpers (swapping client.class.method for client.method)

Update Spanner sample to use client.execute_partition_update instead of client.execute_partitioned_update to avoid NoMethodError.

Update Spanner tests to use timestamp.to_i instead of Time.at(Google::Cloud::Spanner::Data)


tbpg:
I think this is good to submit - want to get tests passing again.

vision/product_search/spec/spec_helper.rb
```diff
@@ -17,12 +17,11 @@
 
 RSpec.configure do |config|
   config.before(:all) do
-    skip("Awaiting a fix in Kokoro project authentication")
     @current_directory = File.expand_path(File.dirname(__FILE__))
     @client = Google::Cloud::Vision::ProductSearch.new
-    @project_id = ENV["GOOGLE_CLOUD_PROJECT"]
+    @project_id = ENV["E2E_GOOGLE_CLOUD_PROJECT"]
```
alixhami: Why is this the project? This should be the same environment variable as other products/languages to create less contributor friction

TheRoyalTnetennba: My thought process was - b/c the product search tests require multiple runs on the same project to pass, I wanted to avoid using the projects provided by gimmeproj.

tbpg: Not sure I understand - can you run it multiple times on the same project regardless? Or is the problem the service account project and test project being the same?

alixhami: Once we fix the way Kokoro runs the tests (service accounts use their own project), I'll have to go in and set up the resources on each project. Because each service account needs to use its own project, I think this actually might create more problems rather than solve them.

TheRoyalTnetennba: This any better? 
+ https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/5795ba73e09e7d5d9ac768cc6df44e55d8219eca/spec/get_service_account.rb
+ https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/5795ba73e09e7d5d9ac768cc6df44e55d8219eca/spec/kokoro-run-all.sh#L54
+ https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/5795ba73e09e7d5d9ac768cc6df44e55d8219eca/vision/product_search/spec/spec_helper.rb#L22

alixhami: I'll leave it to @tbpg to decide what works because he's implementing the kokoro fix. My preference is just that contributors can run the tests with their standard GOOGLE_CLOUD_PROJECT environment variable set

TheRoyalTnetennba: @tbpg It needs to run multiple times b/c some of the tests assume that the created products have been indexed which takes a while https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/329ede052c702f36f377cd66cbbae14af8e76098/vision/product_search/spec/product_search_get_similar_products_spec.rb#L47

alixhami: Yes, I am going to have to manually run the tests for each of the kokoro projects so that the products get indexed. But each service account needs to use its own project or none of the tests will run at all (vision doesn't allow one service account to access resources in another project). They're all currently being skipped until kokoro is set up to have each service account run the tests with its own project (instead of all service accounts using one shared project): https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/329ede052c702f36f377cd66cbbae14af8e76098/vision/product_search/spec/spec_helper.rb#L20

PR #385, opened 2019-01-11
Monitoring  - alerts examples

jiren:
Issue: #359

- monitoring_alert_backup_policies
- monitoring_alert_restore_policies
- monitoring_alert_replace_channels
- monitoring_alert_enable_policies
- monitoring_alert_enable_channel
- monitoring_alert_list_channels
- monitoring_alert_delete_channel
- monitoring_alert_create_channel
- monitoring_alert_list_policies
- monitoring_alert_create_policy
- monitoring_alert_update_channel


sduskis:
@SurferJeffAtGoogle, this is available for review.

monitoring/alerts.rb
```diff
@@ -0,0 +1,216 @@
+require "json"
+require "google/cloud/monitoring"
+require "google/cloud/monitoring/v3/alert_policy_service_client"
```
TheRoyalTnetennba: move requires into the relevant region tags

TheRoyalTnetennba:
Please rebase onto master and make sure `bundle exec rubocop` passes without issue

monitoring/alerts.rb
```diff
@@ -0,0 +1,216 @@
...
+require "google/cloud/monitoring/v3/notification_channel_service_client"
+
+def list_alert_policies project_id:
+  # [START monitoring_alert_list_policies]
+  client = Google::Cloud::Monitoring::AlertPolicy.new
```
TheRoyalTnetennba: If a variable is not defined within a region tag, please add a commented line defining it with a generic value i.e. `# project_id = "your_project_id"`

monitoring/alerts.rb
```diff
@@ -0,0 +1,216 @@
...
+  # [END monitoring_alert_enable_policies]
+end
+
+def list_notification_channels project_id:
+  # [START monitoring_alert_list_channels]
+  client = Google::Cloud::Monitoring::NotificationChannel.new
+  project_name = Google::Cloud::Monitoring::V3::NotificationChannelServiceClient.project_path project_id
+  channels = client.list_notification_channels project_name
+  channels.each do |channel|
+    p channel
```
TheRoyalTnetennba: Would it be possible to puts an identifying attribute of the channel, rather than displaying the entire object? i.e. channel.name?

monitoring/alerts.rb
```diff
@@ -0,0 +1,216 @@
...
+  policies = client.list_alert_policies project_name, filter: filter
+
+  policies.each do |policy|
+    if policy.enabled.value == enable
+      p "Policy #{policy.name} is already #{policy.enabled.value ? "enabled" : "disabled"}"
+    else
+      policy.enabled.value = enable
+      update_mask = Google::Protobuf::FieldMask.new paths: ["enabled"]
+      # client.update_alert_policy policy, update_mask
+      p "#{enable ? "Enabled" : "Disabled"} #{policy.name}"
```
TheRoyalTnetennba: Nit: could we do `puts` instead of `p` for strings?

monitoring/alerts.rb
```diff
@@ -0,0 +1,216 @@
...
+  File.write("backup.json", JSON.pretty_generate(record))
+  puts "Backed up alert policies and notification channels to 'backup.json'."
+  # [END monitoring_alert_backup_policies]
+end
+
+def restore project_id:
+  # [START monitoring_alert_restore_policies]
+  # [START monitoring_alert_create_policy]
+  # [START monitoring_alert_create_channel]
+  # [START monitoring_alert_update_channel]
```
TheRoyalTnetennba: I know we wouldn't be the only language to do it this way, but it seems weird having a bunch of region tags start and end  in sync. Is there a reasonable way to break it up, making each region tag cover just the specific snippet it corresponds to?

monitoring/alerts.rb
jiren: moved to respective examples.

monitoring/alerts.rb
jiren: added project comments.

monitoring/alerts.rb
jiren: done

monitoring/alerts.rb
jiren: done

monitoring/alerts.rb
jiren: Working on it.

jiren: @TheRoyalTnetennba added separate examples

SurferJeffAtGoogle:
Closing due to inactivity.  Please re-open if you'd still like to merge.

PR #389, opened 2019-01-16
Add Scheduler Sample

averikitsch:


scheduler/spec/scheduler_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect(last_response.body).to include("Hello World!")
+  end
+
+  it "posts to /log_payload" do
+    post "/log_payload", "Hello"
+    expect(last_response.body).to include("Printed job payload")
+  end
+  
+  it "can create and delete a job" do
+    response = create_job(GOOGLE_CLOUD_PROJECT, LOCATION_ID, "my-service")
```
TheRoyalTnetennba: Is it possible to randomize the service_id, so there wouldn't be any issues w/ parallel CI tasks? Otherwise LGTM

averikitsch: The service_id refers to the App Engine service for routing purposes. It doesn't actually have to exist to create the job, so this is just a placeholder. I don't think there should be any issues because the test creates a queue and passes that name to be deleted.

PR #397, opened 2019-02-05
Speech API multichannel

dizcology:
Add sample for Speech API's multichannel feature.  Based on https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/speech/cloud-client/beta_snippets.py#L178


speech/speech_samples.rb
```diff
@@ -362,6 +362,71 @@ def speech_transcribe_model_selection file_path: nil, model: nil
...
+  }
+
+  audio_file = File.binread audio_file_path
+  audio      = { content: audio_file }
+
+  response = speech.recognize config, audio
+
+  results = response.results
+
+  results.each_with_index do |result, i|
```
beccasaurus: These are a slightly different style than the other Ruby samples _(re: how they go through results/alternatives)_, but I like this a lot (and this is more _correct_, looking at the first alternative of each result) 👍

+1

speech/speech_samples.rb
```diff
@@ -401,6 +470,7 @@ def speech_transcribe_model_selection file_path: nil, model: nil
   auto_punctuation          <filename> Detects speech in a local audio file, including automatic punctuation in the transcript.
   enhanced_model            <filename> Detects speech in a local audio file, using a model enhanced for phone call audio.
   model_selection           <filename> Detects speech in a local file, using a specific model.
+  multichannel              <filename> Detects speech in separate channels in a local file.
```
beccasaurus: Missing `multichannel_gcs`

dizcology: done

PR #401, opened 2019-02-08
Add Basic Rubocop Rules

grant:
See b/71750459.


grant:
Graham said there's a project to add gem-installable collection of rubocop rules for use across the different github repos. Closing in favor of that.

PR #406, opened 2019-03-04
Video GA text object

dizcology:
samples for video intelligence API:

- text detection
- object tracking


video/spec/video_samples_spec.rb
```diff
@@ -78,5 +78,21 @@
       /cultural/
     ).to_stdout
   end
+
+  it "can detect texts from a gcs file" do
```
beccasaurus: Only 2 of the 4 samples are tested ~ could you add the local ones?

Ideally the file in cloud-samples-data used for testing isn't enormous ~ if you look at the existing samples, we simply pull the bytes down from cloud-samples-data to execute local file samples

dizcology: done.

PR #409, opened 2019-03-16
storage: add bucket policy only samples

frankyn:
Samples for bucket policy only:
https://cloud.google.com/storage/docs/using-bucket-policy-only


frankyn:
Thanks @TheRoyalTnetennba!

PR #415, opened 2019-05-03
Add Flex websockets sample

andrewsg:


andrewsg:
I do not understand why the presubmit test is trying to delete an appengine version and failing, then trying to kill a process that doesn't exist. Tests appear to be passing though.

andrewsg:
Yes. However, a global variable is central to the sample (the replacement for a global variable here is to use memcache, but setting that up is out of scope. documentation addresses this.)

tbpg:
I wonder if there's a way to either disable the lint rule or move the variable to the app instance (`Sinatra::Base`)? Looks like that's the reason 2.4 failed.

andrewsg:
Okay, I believe I've addressed this by disabling this check in Rubocop for this file only and by explaining the global in detail in comments.

tbpg:
Adding @TheRoyalTnetennba for style changes. I'm still looking at the E2E failure.

tbpg:
Looks like the style check still failed:

```
appengine/websockets/app.rb:15:1: W: Lint/MissingCopEnableDirective: Re-enable Style/GlobalVars cop with # rubocop:enable after disabling it.
# rubocop:disable Style/GlobalVars
^
```

Line 221 of https://source.cloud.google.com/results/invocations/ce4d0a4f-7ccb-48e2-8843-1fde866afe5f/targets/cloud-devrel%2Fruby%2Fruby-docs-samples%2Fpresubmit%2Fruby-2.4/log.

andrewsg:
Sorry, I wasn't sure how to run rubocop locally with the correct settings and I was sure it would work based on the article I read. I added a re-enable bit at the end, hopefully this will work now. I'd add a Robocop meme or something here but I don't want to tempt fate.

tbpg:
Tests are passing with the latest style updates. The E2E issue is not causing a failure, which is OK, but we should fix it separately (will need to update `spec/e2e.rb`.

Should wait for style review from @TheRoyalTnetennba before merging.

PR #420, opened 2019-05-15
Update task comments

averikitsch:


averikitsch:
I'm currently creating the correct queue for the tests to pass.

PR #425, opened 2019-05-30
spec: don't print secrets

tbpg:


tbpg:
Thanks! I re-added the variable check.

PR #430, opened 2019-06-07
Improve Cloud Spanner samples that transfer marketing budget

hegemonic:
The samples that transfer part of an album's marketing budget had some issues:

+ `read_write_transaction`: Checked the existing budget for `second_album` against an arbitrary number rather than the transfer amount. Copied and pasted the transfer amount instead of using a variable.
+ `write_with_transaction_using_dml`: Transferred money from `first_album` to `second_album` instead of the other way around. Checked the existing budget for `first_album` against an arbitrary number rather than the transfer amount. Copied and pasted the transfer amount instead of using a variable.

This change fixes these issues and updates the tests accordingly.

@jsimonweb PTAL

### DIFF
--- a/spanner/spanner_samples.rb
+++ b/spanner/spanner_samples.rb
@@ -503,17 +503,18 @@ def read_write_transaction project_id:, instance_id:, database_id:
 
   require "google/cloud/spanner"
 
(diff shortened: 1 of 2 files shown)


hegemonic:
Friendly ping for @jsimonweb and the repo owners.

PR #433, opened 2019-06-14
BigQuery Snippets: Use autodetected location

grant:
Use autodetected location.

> In the example provided, it is written that The location must be specified, but the example works even without setting location for datasets located in US and asia-northeast1. Whether specifying location for an upload job is an absolute requirement or not is affecting an application I'm working on. Can you provide details?

See [b/116968956](http://b/116968956)

### DIFF
--- a/bigquery/snippets/load_from_file.rb
+++ b/bigquery/snippets/load_from_file.rb
@@ -24,8 +24,6 @@ def load_from_file(dataset_id = "your_dataset_id",
   load_job = dataset.load_job table_id, file_path do |config|
     config.skip_leading = 1
     config.autodetect   = true
-    # Must match the destination dataset location.
-    config.location     = "US"
   end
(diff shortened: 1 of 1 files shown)


tswast:
Can you add a snippet such as https://github.com/googleapis/google-cloud-python/pull/8422 that does show how to override this?

grant:
@tswast I can't add a new snippet. That's not in scope of the sample I'm trying to standardize, fixing the attached bug.
I'm updating this snippet in a few languages.

I can change this PR to help fix the bug in some other way, but would like explicit instructions on how to do that.

tswast:
Per the bug, if you aren't adding a new snippet (location autodetection is a new feature, after all), please add an explanation comment as we did in Java. https://github.com/googleapis/google-cloud-java/pull/5475

grant:
Like the Java snippet, I added a note that the `location` is inferred here.

PR #434, opened 2019-06-28
Add Datatypes examples to Spanner sample.

jsimonweb:


spanner/spanner_samples.rb
```diff
@@ -968,6 +968,267 @@ def update_using_batch_dml project_id:, instance_id:, database_id:
...
+  require "google/cloud/spanner"
+
+  spanner = Google::Cloud::Spanner.new project: project_id
+  client  = spanner.client instance_id, database_id
+
+  example_float = 0.8
+  sql_query = "SELECT VenueId, VenueName, PopularityScore FROM Venues
+               WHERE PopularityScore > @popularity_score"
+
+  params      = { popularity_score: example_float }
```
frankyn: Why not put the value of `example_float` in place?

jsimonweb: The goal of this addition to the sample is to show how to create a variable set which is set to the value of given datatype and then use it as a parameter in a Cloud Spanner query. In Ruby this very straightforward :), so I can update the PR if you prefer the parameter values in place.

frankyn: Apologies for the delay. I would recommend doing in place parameter values instead.

Please update.

frankyn:
Spanner tests only run the latest version per [documentation](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/1c21dc6d606e3dc2f182d55fdf3ae3464f45962d/spec/kokoro-run-all.sh#L6). That's why only one passes.

@jsimonweb, I'll find time to sync with you today. The tests looks okay to me but I'm probably missing something.

PR #435, opened 2019-07-01
Container analysis samples

DanSanche:
added v1 of the container analysis samples. This is my first experience with Ruby, so please be thorough


DanSanche:
I looks like the tests are failing because the projects don't have the API enabled. What's the process for getting that set up?

tbpg:
I enabled the API for all of the projects.

container-registry/container-analysis/sample.rb
```diff
@@ -0,0 +1,346 @@
...
+  puts response.name
+  # [END containeranalysis_get_occurrence]
+  response
+end
+
+def get_occurrences_for_image resource_url:, project_id:
+  # [START containeranalysis_occurrences_for_image]
+  # Initialize the client
+  # resource_url = "The URL of the resource associated with the occurrence.
+  #                 eg. https://gcr.io/project/image@sha256:123"
```
grant: `e.g.` (for exempli gratia)

container-registry/container-analysis/sample.rb
```diff
@@ -0,0 +1,346 @@
...
+  response = client.create_occurrence formatted_project, occurrence
+  puts response.name
+  # [END containeranalysis_create_occurrence]
+  response
+end
+
+def delete_occurrence occurrence_id:, project_id:
+  # [START containeranalysis_delete_occurrence]
+  # occurrence_id = "The API-generated identifier associated with the
+  #                  occurrence"
```
grant: I think some of these comments could be put on a single line?

container-registry/container-analysis/sample.rb
```diff
@@ -0,0 +1,346 @@
...
+  # Initialize the client
+  client = Grafeas.new
+
+  formatted_parent = Grafeas::V1::GrafeasClient.project_path project_id
+  filter = "resourceUrl = \"#{resource_url}\""
+  count = 0
+  client.list_occurrences(formatted_parent, filter: filter).each do |occurrence|
+    # Process occurrence here
+    puts occurrence
+    count += 1
```
grant: Using `count = 0` and then `count += 1` seems like it could be improved.

Could we have something like:
```rb
occurrences = client.list_occurrences(formatted_parent, filter: filter)
occurrences.each {|occurrence| puts occurrence }
puts "Found #{occurrences.count} occurrences"
```

Here and other places.

container-registry/container-analysis/spec/sample_spec.rb
```diff
@@ -0,0 +1,287 @@
...
+          affected_cpe_uri: 'your-uri-here',
+          affected_package: 'your-package-here',
+          min_affected_version: { kind: Grafeas::V1::Version::VersionKind::MINIMUM },
+          fixed_version: { kind: Grafeas::V1::Version::VersionKind::MAXIMUM }
+        ]
+      }
+    }
+    client.create_occurrence(formatted_project, occurrence)
+
+    # try again
```
grant: This block could use more descriptive comments and variables (`try`, `c`)

For example
```rb
# Retry until we find an image with a high severity vulnerability
retry_count = 0
vulnerability_count = 0
```

container-registry/container-analysis/sample.rb
DanSanche: Not without going over 80 characters. Isn't that a hard limit for our ruby samples?

grant: Yes, but perhaps we could simplify the wording in some comments. Example:

> The API-generated id associated with this occurrence

container-registry/container-analysis/sample.rb
DanSanche: The point of this sample is to show how you would retrieve a list of occurrences and do some generic work on each one. Getting the count of the results isn't actually necessary in itself, just the iteration part.

We could completely remove the counting aspect, but I think it's useful to have that part consistent across all languages

DanSanche:
@grant Thanks for looking it over! I don't think there's going to be anyone who has more experience with the service to be honest. I mostly just need a Ruby style review since these samples were ported from other languages, but I can wait for another reviewer if you don't have the bandwidth right now

container-registry/container-analysis/sample.rb
grant: OK, but can we just use `each_with_index` then instead? We can still have a  multiline for-loop body (maybe with a comment) if we expect some generic work to be done.

container-registry/container-analysis/README.md
```diff
@@ -0,0 +1,48 @@
...
+These samples show how to use the [Google Cloud Container Analysis Client Library](https://cloud.google.com/container-registry/docs/reference/libraries).
+
+## Build and Run
+1.  **Enable APIs** - [Enable the KMS API](https://console.cloud.google.com/flows/enableapi?apiid=containeranalysis.googleapis.com)
+    and create a new project or select an existing project.
+1.  **Install and Initialize Cloud SDK**
+    Follow instructions from the available [quickstarts](https://cloud.google.com/sdk/docs/quickstarts)
+1.  **Clone the repo** and cd into this directory
+
+```
```
grant: These code blocks need to be indented 4 spaces or 1 tab for the GitHub markdown parser to continue the ordered list enumeration.

Otherwise, they'll just be 1., 1., 1.

See the md preview to see what I mean.

container-registry/container-analysis/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Description
+
+These samples show how to use the [Google Cloud Container Analysis Client Library](https://cloud.google.com/container-registry/docs/reference/libraries).
+
+## Build and Run
+1.  **Enable APIs** - [Enable the KMS API](https://console.cloud.google.com/flows/enableapi?apiid=containeranalysis.googleapis.com)
+    and create a new project or select an existing project.
+1.  **Install and Initialize Cloud SDK**
+    Follow instructions from the available [quickstarts](https://cloud.google.com/sdk/docs/quickstarts)
```
grant: Add a `-` on line 24 so that the list item subtext is not on the same line as the list item text.

container-registry/container-analysis/spec/sample_spec.rb
grant: Still need a better comment.

container-registry/container-analysis/README.md
DanSanche: hmm interesting. I copied the README from [kms](https://github.com/GoogleCloudPlatform/ruby-docs-samples/tree/master/kms), but it looks like they have the same issues

container-registry/container-analysis/sample.rb
DanSanche: What is the benefit of using `each_with_index` rather than `each`?

grant: So you don't need to define `count` in the loop body. Do you think it's cleaner?

DanSanche: hmm I don't think so, I feel like using the index in that way would just complicate what would otherwise be a simple example. Is this pretty unconventional for Ruby?

I'm not completely opposed to removing the whole "count" part of the sample entirely and just printing the occurrence in each loop, but then it would be inconsistent with other languages

PR #436, opened 2019-07-17
[Cloud Tasks] Update client library version

averikitsch:


tasks/create_http_task.rb
```diff
@@ -24,7 +24,7 @@
 # @param [Integer] seconds The delay, in seconds, to process your task.
 def create_http_task project_id, location_id, queue_id, url, payload: nil, seconds: nil
   # Instantiates a client.
-  client = Google::Cloud::Tasks.new
+  client = Google::Cloud::Tasks::V2beta3.new
```
TheRoyalTnetennba: Why did you choose v2beta3 over the default v2?

averikitsch: I'm slightly concerned that the tests were passing without the v2beta3 because the http_request in this sample only works with v2beta3

averikitsch:
ping @TheRoyalTnetennba

tasks/create_http_task.rb
TheRoyalTnetennba: Ok. Could you change the initialization to
```suggestion
  client = Google::Cloud::Tasks.new version: :v2beta3
```
and revert the require?

averikitsch: done. Thanks!

PR #444, opened 2019-08-15
docs(samples): Add `fs_update_document_increment` snippet

BenWhitehead:


firestore/add_data.rb
```diff
@@ -168,13 +168,38 @@ def update_server_timestamp project_id:
   # project_id = "Your Google Cloud Project ID"
 
   firestore = Google::Cloud::Firestore.new project_id: project_id
+  firestore.doc("cities/new-city-id").set(
```
frankyn: Is this an expected change?

firestore/add_data.rb
```diff
@@ -168,13 +168,38 @@ def update_server_timestamp project_id:
...
+  firestore.doc("cities/dc").set(
+      name:       "Washington D.C.",
+      state:      nil,
+      country:    "USA",
+      capital:    true,
+      population: 680_000
+  )
+  # [START fs_update_document_increment]
+  city_ref = firestore.doc "cities/dc"
+  city_ref.update population: Google::Cloud::Firestore::FieldValue.increment(50)
```
frankyn: is there a shorthand for `Google::Cloud::Firestore::FieldValue.increment(50)`?

firestore/add_data.rb
```diff
@@ -168,13 +168,38 @@ def update_server_timestamp project_id:
...
   city_ref.update timestamp: firestore.field_server_time
   # [END fs_update_server_timestamp]
   puts "Updated the timestamp field of the new-city-id document in the cities collection."
 end
 
+def update_document_increment project_id:
+  # project_id = "Your Google Cloud Project ID"
+
+  firestore = Google::Cloud::Firestore.new project_id: project_id
+  firestore.doc("cities/dc").set(
```
frankyn: I'd expect `cities/DC` to follow existing usage.

firestore/add_data.rb
frankyn: After speaking with @BenWhitehead offline, it looks like this is necessary but not sure how this was first merged without passing CI. Question for another day so long as tests pass.

PR #449, opened 2019-09-16
add cloudsql example

TheRoyalTnetennba:
Adds a ruby version of https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/cloud-sql/mysql/sqlalchemy.

cc @kurtisvg


cloud-sql/mysql/activerecord/app.yaml
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+entrypoint: bundle exec rackup --port $PORT
+env: flex
+runtime: ruby
+
+env_variables:
+  SECRET_KEY_BASE: [SECRET_KEY]
```
kurtisvg: Should this set these variables too?
```yaml
export INSTANCE_CONNECTION_NAME='<MY-PROJECT>:<INSTANCE-REGION>:<INSTANCE-NAME>'
export MYSQL_USER='my-db-user'
export MYSQL_PASS='my-db-pass'
export MYSQL_DATABASE='my_db'
```

cloud-sql/mysql/activerecord/config/database.yml
```diff
@@ -0,0 +1,46 @@
...
+  database: db/development.sqlite3
+
+# Warning: The database defined as "test" will be erased and
+# re-generated from your development database when you run "rake".
+# Do not set this db to the same as development or production.
+test:
+  <<: *default
+  database: db/development.sqlite3
+
+# [START production]
```
kurtisvg: Can we remove these region tags?

cloud-sql/mysql/activerecord/config/database.yml
```diff
@@ -0,0 +1,46 @@
...
+#   gem 'sqlite3'
+#
+default: &default
+  adapter: sqlite3
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  timeout: 5000
+
+development:
+  <<: *default
+  database: db/development.sqlite3
```
kurtisvg: Is this correct? Shouldn't the development be using the database connection via the proxy?

cloud-sql/mysql/activerecord/config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
frankyn: Does this work right now? 
You may want to drop ActionCable from the sample.

kurtisvg:
Any status update on this PR? It's tied to a Q3 OKR, so it would be great if we can get it updated and merged before the end of the week.

cloud-sql/mysql/activerecord/config/database.yml
TheRoyalTnetennba: Totally up to you. I set it up to use sqlite3 for dev and cloudsql-mysql for production. The README instructions account for this. Happy to change it if you'd prefer.

kurtisvg: I think the proxy may be preferred, as it will help users understand how to configure to connect from non-appengine locations.

kurtisvg:
This LGTM pending region tags so we can reference from the docs.

kurtisvg:
It looks like we are still missing quite a few region tags needed in this sample. 

There should be 6 region tags ([devrel/samples/1733](https://devrel.corp.google.com/samples/1733)). For more explicit info on what these snippets should cover, see go/csql-connection-samples or the [Managing Database connections page](https://cloud.google.com/sql/docs/mysql/manage-connections) where they will be used.

kurtisvg:
I see everything except `
cloud_sql_mysql_activerecord_lifetime` - can ActiveRecord limit the lifetime of the underlying connection? If not we can leave it out like we do for [Node](https://cloud.google.com/sql/docs/mysql/manage-connections#duration).

TheRoyalTnetennba:
I can't find a way to set the connection lifetime.

PR #452, opened 2019-09-20
Update the ruby pubsub sample to work in App Engine standard env as well

darrencarlton:


appengine/pubsub/app.yaml
```diff
@@ -16,10 +16,10 @@ runtime: ruby
 env: flex
 entrypoint: bundle exec ruby app.rb
 
-# [START gae_flex_pubsub_yaml]
+# [START gae_pubsub_yaml]
```
averikitsch: Do you want to change this region_tag? You'll have to update all the docs.

appengine/pubsub/app.standard.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright 2019 Google, Inc
```
averikitsch:
```suggestion
# Copyright 2019 Google LLC
```

appengine/pubsub/app.standard.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+runtime: ruby25
+entrypoint: bundle exec ruby app.rb
+
```
averikitsch: Does this need an instance class? https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/appengine/standard-hello_world/app.yaml#L18

appengine/pubsub/app.yaml
darrencarlton: Yes, having one tag for both flex and standard simplifies the doc tagging.

appengine/pubsub/app.standard.yaml
darrencarlton: Sure, let's add F2 as in the other sample, It isn't necessary, but seems like a reasonable setting.
I'll test before committing the PR

darrencarlton: app deploys and runs with instance_class: F2

darrencarlton:
Hi Averi:

I don't have permission to merge the pull request. Can you do the merge?

darrencarlton:
Thanks, Tyler

PR #466, opened 2019-11-05
demo ForceEqualSignAlignment

TheRoyalTnetennba:
Experimenting with forcing equal sign alignment. If both `Layout/ExtraSpacing/ForceEqualSignAlignment` and `Layout/SpaceAroundOperators` are enabled, rubocop gets caught in an infinite loop.

Mixed feelings about this.

### DIFF
--- a/jobs/V3/filter_search_sample.rb
+++ b/jobs/V3/filter_search_sample.rb
@@ -21,25 +21,25 @@ def job_discovery_basic_keyword_search project_id:, company_name:, query:
   require "google/apis/jobs_v3"
 
   # Instantiate the client
-  jobs = Google::Apis::JobsV3
-  talent_solution_client = jobs::CloudTalentSolutionService.new
+  jobs                                 = Google::Apis::JobsV3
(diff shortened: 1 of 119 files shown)


jobs/V3/filter_search_sample.rb
```diff
@@ -21,25 +21,25 @@ def job_discovery_basic_keyword_search project_id:, company_name:, query:
...
+                                                                   domain:     "www.google.com"
 
   # Perform a search for analyst  related jobs
-  job_query = jobs::JobQuery.new query: query
-  job_query.company_names = [company_name] unless company_name.nil?
-  search_jobs_request = jobs::SearchJobsRequest.new request_metadata: request_metadata,
-                                                    job_query:        job_query,
-                                                    search_mode:      "JOB_SEARCH"
-  search_jobs_response = talent_solution_client.search_jobs project_id, search_jobs_request
+  job_query                            = jobs::JobQuery.new query: query
```
TheRoyalTnetennba: An example of where I feel it's overly aggressive. Would prefer the empty line + comments separating these sections would reset the alignment

PR #469, opened 2019-11-23
Add AutoML Samples

blowmage:
This is Ruby implementation of the Python samples from GoogleCloudPlatform/python-docs-samples#2481.

I don't believe the Kokoro build will succeed for this PR until the CGP project has all the AutoML services properly configured and Kokoro is configured with the environment variables needed to run the tests.

### Caveats

I have been unable to create a Vision Classification Dataset and Model, which means that those samples are not under tests and may be incorrect. Those tests are currently skipped until a Dataset and Model can be used to test against.

All but one of the `*_create_model` files from the original python PR were not used in the tests. This PR does add tests for those samples to make sure they work, although the long running operation to create the model is quickly cancelled. Although one test is skipped because it is not able to be cancelled.

The current AutoML Ruby client does not expose the `get_operation` or `list_operations` methods. I am working on a PR to add this to the Ruby AutoML client, but we will have to wait for those changes to be merged and released before those samples can be implemented in this repo.


blowmage:
The `GETTING_STARTED.md` file includes the directions for setting up a project to run these samples, including what is needed to run these samples on Kokoro. All the services need the quickstart completed, and the dataset and model ids added as environment variables.

blowmage:
The `get_operation` and `list_operations` RPCs were added in https://github.com/googleapis/google-cloud-ruby/pull/4458, but they have not yet been released.

PR #470, opened 2019-11-26
Add CLI support to translate_v3_translate_text

blowmage:
This PR tweaks the existing translate samples in a couple ways. First, it removes the mocks from the specs. Second, it adds code to each sample to allow it to be called directly. Third, it fixes a complaint I have about the other samples that are being generated, namely that the variables that are being set by the test or by calling the file directly are commented out in the sample code. IMO the sample should not direct users to uncomment and change the values in the code to run, but to change the values only. This PR implements this by stopping and starting the sample tags to set the actual values to be used.


TheRoyalTnetennba:
@blowmage Trying to clean up PRs. Is this ready/something you intend to work on?

PR #476, opened 2019-12-17
[Storage] IAM condition sample updates

frankyn:


frankyn:
@TheRoyalTnetennba could you help review this PR?

storage/iam.rb
```diff
@@ -62,14 +68,56 @@ def remove_bucket_iam_member project_id:, bucket_name:, role:, member:
...
-  bucket.policy do |policy|
-    policy.remove role, member
+  bucket.policy requested_policy_version: 3 do |policy|
+    policy.bindings.each do |binding|
+      if binding.role == role
+        if binding.members.count > 1
+          # At least 2 members exist in the binding
+          binding.members.delete member
+        else
+          # Last member in the list, delete binding from policy.
```
broady: Is this necessary? Won't the server clean up this binding? What if it has a condition but no members? (Is that why you need to remove the binding?)

storage/iam.rb
```diff
@@ -62,14 +68,56 @@ def remove_bucket_iam_member project_id:, bucket_name:, role:, member:
...
+          policy.bindings.remove binding
```
broady: Is it valid to remove while iterating over the bindings list?

storage/iam.rb
frankyn: If there are no members in the array IAM will fail paraphrasing "at least 1 member is required".

storage/iam.rb
frankyn: :man_facepalming: good catch will fix.

storage/iam.rb
frankyn: Pending fix in the service this will go away. Until then the sample update is blocked.

frankyn:
@GoogleCloudPlatform/yoshi-ruby anyone have a moment to review these changes from a Ruby perspective?

PR #486, opened 2020-01-28
[bigtable] read and filter snippets

billyjacobson:


bigtable/filter_samples.rb
```diff
@@ -0,0 +1,217 @@
...
+# Import google bigtable client lib
+require "google/cloud/bigtable"
+
+# [START bigtable_filters_limit_row_sample]
+def filter_limit_row_sample project_id, instance_id, table_id
+  filter = Google::Cloud::Bigtable::RowFilter.sample 0.75
+  read_with_filter project_id, instance_id, table_id, filter
+end
+
+# [END bigtable_filters_limit_row_sample]
```
TheRoyalTnetennba: nit: can you remove the line between "end" and "# [END" and add one between "END" and "START"

bigtable/spec/filter_samples_spec.rb
```diff
@@ -0,0 +1,613 @@
...
+\tos_build: PQ2A.190406.000 @#{@timestamp}
+
+"
+  end
+
+  it 'filter_limit_col_family_regex' do
+    output = capture do
+      filter_limit_col_family_regex @project_id, @instance_id, @table_id
+    end
+    expect(output).to match "Reading data for phone#4c410523#20190501:
```
TheRoyalTnetennba: Nit: For multi-line strings, could you do 
`<<~OUTPUT`
`        Reading data for phone`...
`        tos_build: PQ2A.190406.000 @#{@timestamp}`
`OUTPUT`

bigtable/filter_samples.rb
```diff
@@ -0,0 +1,217 @@
...
+
+
+# Import google bigtable client lib
+require "google/cloud/bigtable"
+
+# [START bigtable_filters_limit_row_sample]
+def filter_limit_row_sample project_id, instance_id, table_id
+  filter = Google::Cloud::Bigtable::RowFilter.sample 0.75
+  read_with_filter project_id, instance_id, table_id, filter
+end
```
TheRoyalTnetennba: can you move the `START` to after the `def` and the `END` before the `end`

bigtable/filter_samples.rb
```diff
@@ -0,0 +1,217 @@
...
+  )
+                                             .on_match(Google::Cloud::Bigtable::RowFilter.label("passed-filter"))
+                                             .otherwise(Google::Cloud::Bigtable::RowFilter.label("filtered-out"))
+  read_with_filter project_id, instance_id, table_id, filter
+end
+
+# [END bigtable_filters_composing_condition]
+
+
+# [START bigtable_filters_limit_row_sample]
```
TheRoyalTnetennba: Why are these snippets included in so many region tags? Do they do everything the region tag implies they do?

bigtable/filter_samples.rb
```diff
@@ -0,0 +1,217 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Import google bigtable client lib
+require "google/cloud/bigtable"
```
TheRoyalTnetennba: include the require in the individual snippets

bigtable/filter_samples.rb
billyjacobson: So I'm doing this to avoid duplicating the code to read data with a filter and print each row. By using multiple region tags, it will let me include that part in each part of the documentation. The double region tags will get excluded with a regex in the includecode for the documentation.

bigtable/filter_samples.rb
billyjacobson: I can, but then the indentation will look off in the documentation

TheRoyalTnetennba: It should be fine if the region tags are also indented. https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/f3ef01233f51b2df642126bacff69268e69101e7/vision/detect_crop_hints.rb#L18 & https://cloud.google.com/vision/docs/detecting-crop-hints#vision_crop_hint_detection-ruby

billyjacobson: Wow awesome, didn't know it works like that.

bigtable/filter_samples.rb
```diff
@@ -0,0 +1,252 @@
...
+  # [START bigtable_filters_composing_chain]
+  filter = Google::Cloud::Bigtable::RowFilter.chain.cells_per_column(1).family("cell_plan")
+  read_with_filter project_id, instance_id, table_id, filter
+  # [END bigtable_filters_composing_chain]
+end
+
+def filter_composing_interleave project_id, instance_id, table_id
+  filter = Google::Cloud::Bigtable::RowFilter.interleave.value("true").qualifier("os_build")
+  read_with_filter project_id, instance_id, table_id, filter
+  # [END bigtable_filters_composing_interleave]
```
TheRoyalTnetennba: Either this is missing the "START" tag or I am

billyjacobson: fixed

billyjacobson:
> LGTM, aside from the comment. Assuming the spacing looks good once all the region tags are put together, i.e. the require statement for bigtable_filters_limit_row_sample fits into the rest of the snippet well.

thank you! sorry for the delays with the back and forth, was working on perf

TheRoyalTnetennba:
Thanks for doing this

PR #493, opened 2020-02-18
(cloud talent) Updating Ruby samples for v4beta1 Cloud Talent API; out of date

hilts-vaughan:
These are being used in production documentation: https://cloud.google.com/talent-solution/job-search/docs/histogram#job_search_histogram_search-ruby

So we should make sure they are correct.


TheRoyalTnetennba:
Thanks for putting this together. Could you add tests for the new stuff?

hilts-vaughan:
Actually, none of it has tests right now and that was the original authors intent. If it's OK with you, I would prefer this PR to focus on fixing our ueerbgafing documentation. Adding tests, assuming this is a manual process, would be a lot of overhead to this PR.

This PR was generated with synthtool.

I can commit to writing a few tests later if it helps but I'd like our customers to be getting right samples first since I feel like tests won't just be rubber stamped and will go through a few rounds (especially since I haven't seriously worked with Ruby in some time!)

PR #498, opened 2020-03-11
spanner: code samples for query options

hengfengli:


skuruppu:
You'll need to wait for googleapis/google-cloud-ruby/pull/5017 to complete the autorelease process before you can merge this in and probably also need to update the [version](https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/bb937374881d5d025c4d8424e95abeab197ec7c3/spanner/Gemfile.lock#L14).

PR #501, opened 2020-03-19
add remove conditional binding sample

frankyn:


storage/iam.rb
```diff
@@ -109,6 +109,49 @@ def add_bucket_conditional_iam_binding bucket_name:, role:, member:, title:, des
...
+        expression:  expression
+      }
+      if (b.role == role) && (b.condition &&
+        b.condition.title == title &&
+        b.condition.description == description &&
+        b.condition.expression == expression)
+        binding_to_remove = b
+      end
+    end
+    if binding_to_remove
```
jkwlui: I would just merge the removal and `puts`:
```ruby
if binding_to_remove
  policy.bindings.remove binding_to_remove
  puts "Conditional Binding was removed."
else
  puts "No matching conditional binding found."
end
```

frankyn: done!

frankyn:
Failure is a flake.

storage/iam.rb
```diff
@@ -109,6 +109,46 @@ def add_bucket_conditional_iam_binding bucket_name:, role:, member:, title:, des
...
+  # description = "Condition Description"
+  # expression  = "Condition Expression"
+
+  require "google/cloud/storage"
+
+  storage = Google::Cloud::Storage.new
+  bucket = storage.bucket bucket_name
+
+  bucket.policy requested_policy_version: 3 do |policy|
+    policy.version = 3
```
quartzmo: Given that there should be an existing binding with a condition, does `version` need to be set here? In the google-cloud-storage [`Bucket#update_policy` example](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-storage/lib/google/cloud/storage/bucket.rb#L1978), `version` is not set but is shown as already being `3`. Is that example incorrect? (I can investigate.)

frankyn: It's best practices to always set version to 3 given the API design. 

Probably should update that in documentation.

quartzmo: Thanks, will do!

PR #504, opened 2020-03-24
add build-cop

TheRoyalTnetennba:
- add rspec_junit_formatter to all Gemfiles that use rspec
- invoke build-cop on all tests (temporarily)

### DIFF
--- a/spec/kokoro-run-all.sh
+++ b/spec/kokoro-run-all.sh
@@ -197,6 +197,6 @@ else
 fi
 
 chmod +x $KOKORO_GFILE_DIR/linux_amd64/buildcop
-$KOKORO_GFILE_DIR/linux_amd64/buildcop --repo GoogleCloudPlatform/ruby-docs-samples
+$KOKORO_GFILE_DIR/linux_amd64/buildcop
 
 exit $EXIT_STATUS
--- a/spec/kokoro-run-all.sh
+++ b/spec/kokoro-run-all.sh
@@ -181,7 +181,7 @@ else
 
     start_time="$(date -u +%s)"
 
-    (bundle update && bundle exec rspec --format RspecJunitFormatter --out sponge_log.xml) || set_failed_status
+    (bundle update && bundle exec rspec --format documentation --format RspecJunitFormatter --out sponge_log.xml | tee sponge_log.log) || set_failed_status
 
     if [[ $E2E = "true" ]]; then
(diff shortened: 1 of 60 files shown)


spec/kokoro-run-all.sh
```diff
@@ -197,4 +196,7 @@ else
   done
 fi
 
+chmod +x $KOKORO_GFILE_DIR/linux_amd64/buildcop
+$KOKORO_GFILE_DIR/linux_amd64/buildcop --repo GoogleCloudPlatform/ruby-docs-samples
```
tbpg: I sent https://github.com/googleapis/repo-automation-bots/pull/414 so you can remove the `-repo` flag.

spec/kokoro-run-all.sh
```diff
@@ -197,4 +196,7 @@ else
   done
 fi
 
+chmod +x $KOKORO_GFILE_DIR/linux_amd64/buildcop
```
tbpg: Think I saw this in the E2E block, so wanted to comment here for when you go to move it back. I think E2E tests only run on one version, and possibly on PRs. So, we should do:

```
if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"system-tests"* ]]; then
  ...
fi
```

spec/kokoro-run-all.sh
```diff
@@ -173,16 +173,15 @@ if [[ $RUN_ALL_TESTS = "1" ]]; then
...
+  for PRODUCT in $SPEC_DIRS; do
     # Run Tests
     echo "[$PRODUCT]"
     export TEST_DIR="$PRODUCT"
     pushd "$REPO_DIRECTORY/$PRODUCT/"
 
     start_time="$(date -u +%s)"
 
-    (bundle update && bundle exec rspec --format documentation) || set_failed_status
+    (bundle update && bundle exec rspec --format RspecJunitFormatter --out sponge_log.xml) || set_failed_status
```
tbpg: Might want to have both formatters so we get nice stdout for humans and the XML file for computers. Tested this locally and it works. Also, if you `tee` to a log file with the same base name, Sponge/Build Status will know what to do with it (once they're uploaded as artifacts).

```suggestion
    (bundle update && bundle exec rspec --format documentation --format RspecJunitFormatter --out sponge_log.xml | tee sponge_log.log) || set_failed_status
```

PR #564, opened 2020-04-30
datastore: convert to minitest

quartzmo:
refs: #479


datastore/Gemfile
```diff
@@ -15,11 +15,10 @@
 source "https://rubygems.org"
 
 gem "google-cloud-datastore"
-gem "sinatra"
```
quartzmo, line 18: @TheRoyalTnetennba Should `google-style` be required in this file? It is in #489 but not in #542.

datastore/Rakefile
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require "rake/testtask"
+
+Rake::TestTask.new "test" do |t|
+  t.test_files = FileList["**/*_test.rb"]
+  t.options = "--junit --junit-filename=sponge_log.xml"
+  t.warning = false
+end
```
quartzmo, line 21: @TheRoyalTnetennba Should there be an `acceptance ` task in this file? There is one in #489 but not in #542.

datastore/spec/quickstart_spec.rb
```diff
@@ -1,55 +0,0 @@
...
-RSpec.configure do |config|
-  # show retry status in spec process
-  config.verbose_retry = true
-  # show exception that triggers a retry if verbose_retry is set to true
-  config.display_try_failure_messages = true
-
-  # set retry count and retry sleep interval to 10 seconds
-  config.default_retry_count = 5
-  config.default_sleep_interval = 10
-end
```
quartzmo, line 28: Does anyone think I should port this retry configuration to minitest? I did not notice it to be needed in several dozen local test runs, so I omitted it in hopes that it was added at a time when the service was having temporary issues.

datastore/Gemfile
TheRoyalTnetennba: I kinda like it in the sample Gemfile b/c I run `rubocop -a` a lot to fix hash alignment etc.

datastore/spec/quickstart_spec.rb
TheRoyalTnetennba: I have a bias towards retries in tests but idk if the client handles that already

datastore/Rakefile
TheRoyalTnetennba: I don't think the acceptance task is needed atm. Wasn't originally sure how the ci for samples was going to look

datastore/Gemfile
quartzmo: Ok I'll add it.

datastore/Rakefile
quartzmo: Good, thanks!

datastore/spec/quickstart_spec.rb
quartzmo: OK if we add it later when we actually experience service failures? I will be happy to.

datastore/Gemfile
quartzmo: Added.

PR #581, opened 2020-05-08
bigtable: convert to minitest

quartzmo:
refs: #479

This PR updates the test framework for these samples without changing the samples except as noted:
* Add `hello_world` wrapper method.
* Add `quickstart` wrapper method.
* Remove unused `project_id` arguments from sample methods.
* Remove the following `TODO` from `tableadmin_spec.rb` and updates the column family usage samples in `tableadmin.rb` to pass against the latest release of `google-cloud-bigtable`.

  ```
  # TODO Update samples to match new client
  ```


quartzmo:
Kokoro CI appears to have failed, but I don't see a link to the log.

bigtable/filter_samples.rb
```diff
@@ -22,147 +20,147 @@
...
   # [END bigtable_filters_limit_col_qualifier_regex]
 end
 
-def filter_limit_col_range project_id, instance_id, table_id
+def filter_limit_col_range instance_id, table_id
   # [START bigtable_filters_limit_col_range]
   range = Google::Cloud::Bigtable::ColumnRange.new("cell_plan").from("data_plan_01gb").to("data_plan_10gb")
   filter = Google::Cloud::Bigtable::RowFilter.column_range range
-  read_with_filter project_id, instance_id, table_id, filter
+  read_with_filter instance_id, table_id, filter
```
TheRoyalTnetennba, line 76: Could you add comments describing the variables that are set outside the region tags, similar to https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/8ef68b062146ae22eeaac8671bc0ac1bdacba3ad/servicedirectory/servicedirectory.rb#L17

quartzmo: For storage samples, @JesseLovelace [commented](https://github.com/googleapis/google-cloud-ruby/pull/7823#pullrequestreview-513136492) that a combination of description and example value is preferred. Would you agree for Bigtable?

```
# Your Google Cloud project ID
# project = "your-project-id"
```

TheRoyalTnetennba: The snippet rubric doesn't require the explanation. I'd say add a description if you feel one is needed, otherwise an example value is fine.

quartzmo: OK, I'll just add example values for now:

```
# project = "your-project-id"
```

quartzmo: @TheRoyalTnetennba you wrote:

> Could you add comments describing the variables that are set outside the region tags

But did you mean *inside* the region tags?

quartzmo: @TheRoyalTnetennba I added comments describing the variables wherever possible. In some region tags, the `table` is implicit (the lookup is not shown), so in those I did not add the comments.

TheRoyalTnetennba (approved):

PR #588, opened 2020-05-17
Update inline links to postgres docs

glasnt:
Looks like this was copied from the mysql README without updating the links to the postgres-specific docs.


TheRoyalTnetennba:
Thanks!

PR #613, opened 2020-06-17
fix: update to use new ci project

crwilcox:


crwilcox:
Seems the project has migrated to the new secret, but still saw failures, that seem unrelated. Rerunning.

PR #627, opened 2020-07-07
[Spanner] database operation test - removed strict check assert

jiren:
Fixes: #607

### DIFF
--- a/spanner/spec/spanner_samples_spec.rb
+++ b/spanner/spec/spanner_samples_spec.rb
@@ -1516,7 +1516,7 @@ describe "Google Cloud Spanner API samples" do
     end
 
     expect(captured_output).to match(
-      /Database #{database.database_id} restored from backup is \d+% optimized/
+      /Database \w+ restored from backup is \d+% optimized/
     )
   end
 
--- a/spanner/spec/spanner_samples_spec.rb
+++ b/spanner/spec/spanner_samples_spec.rb
@@ -1516,7 +1516,7 @@ describe "Google Cloud Spanner API samples" do
     end
 
     expect(captured_output).to match(
-      /Database \w+ restored from backup is \d+% optimized/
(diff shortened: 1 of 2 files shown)


spanner/spec/spanner_samples_spec.rb
```diff
@@ -1516,7 +1516,7 @@ def capture
     end
 
     expect(captured_output).to match(
-      /Database #{database.database_id} restored from backup is \d+% optimized/
+      /Database \w+ restored from backup is \d+% optimized/
```
skuruppu, line 1519: I see that in the test output there are results that match this. But do we know if this is always true? Only because I think we clear the databases at the end of the tests so I'm not sure if there will always be a result in this list call.

hengfengli:
I don't think the failed test is related to this change: 

```bash
Logging Samples::list_log_entries#test_0001_lists entries of a log [/tmpfs/src/github/ruby-docs-samples/logging/acceptance/sample_test.rb:126]:
```

jiren:
> I don't think the failed test is related to this change:
> 
> ```shell
>   1) Failure:
> Logging Samples::list_log_entries#test_0001_lists entries of a log [/tmpfs/src/github/ruby-docs-samples/logging/acceptance/sample_test.rb:126]:
> Expected false to be truthy.
> ```

Yes logging test is failing.

spanner/spec/spanner_samples_spec.rb
jiren: @skuruppu Got it. In this case i am thinking to pull message like. `List database operations with optimized database filter found #{jobs.length} jobs` in sample code and check with the `/List database operations with optimized database filter found \d+% jobs/`

skuruppu:
@TheRoyalTnetennba would I please be able to ask you to merge in this PR for us? The failed tests are unrelated and are to do with failures in github/ruby-docs-samples/logging.

PR #647, opened 2020-09-12
chore(functions): replace helloGcs sample with helloGcsGeneric

ace-n:
Per b/135560881, we've since replaced _what used to be_ `helloGCS` with _what used to be_ `helloGCSGeneric` (and is now `helloGCS`) to reduce unnecessary duplication.

@labtopia FYI


TheRoyalTnetennba (approved):

ace-n:
Looks like `functions` tests are passing - @TheRoyalTnetennba can you merge if so? Thanks! 🙂

dazuma (approved):

PR #695, opened 2020-10-27
spanner: commit stats example

jiren:


spanner/spanner_samples.rb
```diff
@@ -1873,6 +1873,28 @@ def set_custom_timeout_and_retry project_id:, instance_id:, database_id:
...
+  spanner = Google::Cloud::Spanner.new project: project_id
+  client  = spanner.client instance_id, database_id
+
+  records = [
+    { SingerId: 1, FirstName: "Marc",     LastName: "Richards" },
+    { SingerId: 2, FirstName: "Catalina", LastName: "Smith"    }
+  ]
+  resp = client.upsert "Singers", records, commit_stats: true
+  puts "Updated data with #{resp.stats.mutation_count} mutations."
+  puts "Commit was delayed by #{resp.stats.overload_delay} due to overloaded servers."
```
skuruppu, line 1893: It might be better to print this only if it's specified. Otherwise, I believe it's null.

Also specify a time unit (I need to do the same in Node.js).

jiren: @skuruppu updated.

hengfengli:
LGTM.

spanner/spanner_samples.rb
```diff
@@ -1873,6 +1873,32 @@ def set_custom_timeout_and_retry project_id:, instance_id:, database_id:
...
+
+  spanner = Google::Cloud::Spanner.new project: project_id
+  client  = spanner.client instance_id, database_id
+
+  records = [
+    { SingerId: 1, FirstName: "Marc",     LastName: "Richards" },
+    { SingerId: 2, FirstName: "Catalina", LastName: "Smith"    }
+  ]
+  commit_options = { return_commit_stats: true }
+  resp = client.upsert "Singers", records, commit_options: commit_options
```
skuruppu, line 1892: Sorry that I didn't notice this before. The reference sample we did had the following logic (Node.js):

```
    const [response] = await albumsTable.upsert(
      [
        {SingerId: '1', AlbumId: '1', MarketingBudget: '200000'},
        {SingerId: '2', AlbumId: '2', MarketingBudget: '400000'},
      ],
      {returnCommitStats: true}
    );
```

It would be better if we could keep the samples consistent. Let me know if there's some conflict with the tests that I didn't foresee.

skuruppu (changes requested):

spanner/spanner_samples.rb
jiren: updated example.

skuruppu (approved):

spanner/spanner_samples.rb
```diff
@@ -1873,6 +1873,32 @@ def set_custom_timeout_and_retry project_id:, instance_id:, database_id:
...
+    { SingerId: 2, AlbumId: 2, MarketingBudget: 400_000 }
+  ]
+  commit_options = { return_commit_stats: true }
+  resp = client.upsert "Albums", records, commit_options: commit_options
+  puts "Updated data with #{resp.stats.mutation_count} mutations."
+
+  if resp.stats.overload_delay
+    puts "Commit was delayed by #{resp.stats.overload_delay} seconds due to overloaded servers."
+  end
+
```
skuruppu, line 1898:
```suggestion
```

Sorry @jiren for the trouble but we were asked to remove this field.

jiren: @skuruppu removed

skuruppu: Thanks a lot @jiren.

skuruppu:
@jiren I've done a release of the Spanner client for Ruby so you should be able to get the tests working. Please merge once the tests pass.

jiren:
> @jiren I've done a release of the Spanner client for Ruby so you should be able to get the tests working. Please merge once the tests pass.

Ok.

Issue #726, opened 2021-01-26
Pub/Sub: distinguish between topic and subscription IDs and names


anguillanneuf:
Samples should refer to the fully qualified topic paths `projects/{}/topics/{}` as topic names and the actual string for the topic as the topic id. 

The same goes for subscriptions. 

For example, this should be `topic_id`. 
https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/b1a13ea51c868af607c4c88e768416f3ab47c615/pubsub/topics.rb#L17

quartzmo:
There is a similar situation with `Subscription`. Should we also change `subscription_name` to `subscription_id` where the value is not the fully qualified subscription path?


https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/b1a13ea51c868af607c4c88e768416f3ab47c615/pubsub/subscriptions.rb#L17

anguillanneuf:
Yes. 

Sorry I didn't look too carefully. By `subscription_name`, do you refer to the fully qualified subscription path? If that's the case, we don't have to change anything. 

I had thought that we were using `subscription_name` to refer to `subscription_id`.

anguillanneuf:
Specifically in the README, 

> Example:
> `bundle exec ruby topics.rb create_topic my-new-topic`
> Topic `my-new-topic` created.

I would have thought that `my-new-topic` refers to a topic ID. If what's expected is a topic name, we can just update the README.

quartzmo:
The design in google-cloud-pubsub is consistent, meaning that both `topic_name` and `subscription_name` parameters are **not** the fully qualified path.

See the API docs examples for [Topic#subscribe](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-pubsub/lib/google/cloud/pubsub/topic.rb#L374).

Also, if you look at the ruby-docs-samples test helper, you can see that the value is not fully qualified:

https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/b1a13ea51c868af607c4c88e768416f3ab47c615/pubsub/acceptance/helper.rb#L25-L27

quartzmo:
We may also want to attempt to change these parameter names in google-cloud-pubsub. I haven't yet looked at how hard that will be to do.

quartzmo:
[...]

PR #739, opened 2021-02-04
fix(pubsub): Rename resource name parameters and variables

quartzmo:
* Rename `topic_name` to `topic_id`.
* Rename `subscription_name` to `subscription_id`.
* Change example values to "your-topic-id" and "your-subscription-id", respectively.

closes: #726


anguillanneuf (approved):
Wow, literally all over the place. 
Thanks for this fix!

PR #761, opened 2021-03-02
samples(spanner): parameterise create backup

thiagotnunes:
Instead of hardcoding the version time in the create backup sample, parameterises it.

### DIFF
--- a/spanner/spanner_samples.rb
+++ b/spanner/spanner_samples.rb
@@ -2065,14 +2065,6 @@ def run_sample arguments
                           database_id: database_id,
                           start_title: arguments.shift,
                           end_title:   arguments.shift
-  elsif commands.eql?("create_backup") && instance_id && database_id && arguments.size >= 2
-    require 'time'
-
-    create_backup project_id: project_id,
-                  instance_id: instance_id,
-                  database_id: database_id,
-                  backup_id: arguments.shift,
-                  version_time: Time.parse(arguments.shift)
   elsif commands.include?(command) && instance_id && database_id
     send command, project_id:  project_id,
                   instance_id: instance_id,
--- a/spanner/spanner_samples.rb
+++ b/spanner/spanner_samples.rb
@@ -1670,12 +1670,13 @@ def write_read_float64_array project_id:, instance_id:, database_id:
   # [END spanner_write_read_float64_array]
 end
 
-def create_backup project_id:, instance_id:, database_id:, backup_id:
+def create_backup project_id:, instance_id:, database_id:, backup_id:, version_time:
   # [START spanner_create_backup]
   # project_id  = "Your Google Cloud project ID"
   # instance_id = "Your Spanner instance ID"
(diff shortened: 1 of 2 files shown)


spanner/spanner_samples.rb
```diff
@@ -1684,7 +1685,6 @@ def create_backup project_id:, instance_id:, database_id:, backup_id:
   instance = spanner.instance instance_id
   database = instance.database database_id
   expire_time = Time.now + 14 * 24 * 3600 # 14 days from now
-  version_time = client.execute("SELECT CURRENT_TIMESTAMP() as timestamp").rows.first[:timestamp]
 
   job = database.create_backup backup_id, expire_time, version_time: version_time
```
jiren, line 1689: @thiagotnunes I think here time is a string then need to parse it.

`Time.parse(version_time)`

thiagotnunes: I don't think so, if we do the following it returns a `PreciseDate`

```
version_time = client.execute("SELECT CURRENT_TIMESTAMP() as timestamp").rows.first[:timestamp]
```

spanner/spanner_samples.rb
```diff
@@ -1670,12 +1670,13 @@ def write_read_float64_array project_id:, instance_id:, database_id:
...
 end
 
-def create_backup project_id:, instance_id:, database_id:, backup_id:
+def create_backup project_id:, instance_id:, database_id:, backup_id:, version_time:
   # [START spanner_create_backup]
   # project_id  = "Your Google Cloud project ID"
   # instance_id = "Your Spanner instance ID"
   # database_id = "Your Spanner database ID"
   # backup_id = "Your Spanner backup ID"
+  # version_time = Time.now - 60 * 60 * 24 # 1 day ago
```
thiagotnunes, line 1679: @jiren We also expect the user to pass in a `Time` here.

spanner/spanner_samples.rb
jiren: Got it. Only concern is that running sample using command line it will not work. I think for this we should mention in documentation.

thiagotnunes: In matter of fact, the `create_backup` can not be run through the command line at the moment. Check https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/spanner/spanner_samples.rb#L2062-L2074, where the `create_backup` requires a `backup_id` (https://github.com/GoogleCloudPlatform/ruby-docs-samples/blob/master/spanner/spanner_samples.rb#L1673). It will fail, because the `backup_id` is never passed in.

I tried adding another branch in the `if... else...` statement, but the cyclomatic complexity gets too high and the build fails. I think we should extract different files for the samples, this way it will be more in line with the other languages. However, I don't think it should be a part of this PR.

jiren: Thanks for clarification.

jiren (approved):

PR #764, opened 2021-03-05
chore: give presubmit and system_tests their own common.cfg

TheRoyalTnetennba:
presubmit and system_tests aren't inheriting common.cfg from .kokoro/, so create common.cfg's for each separately.


dazuma (approved):
Stamping, although I'm wondering why the trampoline images are reverting to the older `.../ruby-2.5` images instead of `.../yoshi-ruby/multi-node`

PR #788, opened 2021-04-22
Add comment about authenticated push JWT token validation

fayssalmartanigcp:
This snippet is used in the Cloud Pub/Sub docs (https://cloud.google.com/pubsub/docs/push#validating_tokens) and many users are not aware that signature verification of the token is not enough, the claim needs to be validated also.

See similar change in the nodejs library: GoogleCloudPlatform/nodejs-docs-samples#2222.


TheRoyalTnetennba (approved):

NivedhaSenthil (approved):

bajajneha27:
docs: Add comment about authenticated push JWT token validation (#788)

PR #803, opened 2021-05-18
docs: use single use transaction for quickstart query

olavloite:
The quickstart simple query example should use a single-use read-only transaction, as that is the most efficient way to execute a single query. The current sample uses a read/write transaction, which is the least efficient way to execute a single query.

### DIFF
--- a/spanner/quickstart.rb
+++ b/spanner/quickstart.rb
@@ -32,11 +32,8 @@ database_id = "my-database"
 database_client = spanner.client instance_id, database_id
 
 # Execute a simple SQL statement
-database_client.transaction do |transaction|
-  results = transaction.execute "SELECT 1"
-
-  results.rows.each do |row|
-    puts row
-  end
+results = database_client.execute_query "SELECT 1"
+results.rows.each do |row|
+  puts row
 end
 # [END spanner_quickstart]


olavloite:
The build error seems unrelated to this change.

@skuruppu Do you know who could review this PR?

hengfengli (approved):
LGTM.

spanner/quickstart.rb
```diff
@@ -32,11 +32,8 @@
 database_client = spanner.client instance_id, database_id
 
 # Execute a simple SQL statement
-database_client.transaction do |transaction|
-  results = transaction.execute "SELECT 1"
-
-  results.rows.each do |row|
-    puts row
-  end
+results = database_client.execute_query "SELECT 1"
```
jiren, line 35: @olavloite This will not use single use transaction. To use single use transaction need to provide `single_use` argument.

i.e 
```
database_client.execute_query "SELECT 1", single_use: { timestamp: Time.now }
```
Ref: https://github.com/googleapis/google-cloud-ruby/blob/809ebea224c9b8c9f002c7809afa537b735602d2/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L375

Valid keys for `single_use`
https://github.com/googleapis/google-cloud-ruby/blob/809ebea224c9b8c9f002c7809afa537b735602d2/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L1732-L1736

olavloite: No, that is not required (as is also the default for the other client libraries):
1. Calling `execute_query` without a `single_use` parameter will cause the method to call the `single_use_transaction` method with a nil parameter [here](https://github.com/googleapis/google-cloud-ruby/blob/809ebea224c9b8c9f002c7809afa537b735602d2/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L382).
2. The `single_use_transaction` method will [return nil](https://github.com/googleapis/google-cloud-ruby/blob/809ebea224c9b8c9f002c7809afa537b735602d2/google-cloud-spanner/lib/google/cloud/spanner/client.rb#L1748) when it is called with a nil parameter.
3. That will cause the `session.execute_query` method to be [called with a nil transaction selector](https://github.com/googleapis/google-cloud-ruby/blob/809ebea224c9b8c9f002c7809afa537b735602d2/google-cloud-spanner/lib/google/cloud/spanner/session.rb#L315).
4. This will in the end call the `ExecuteStreamingSql` RPC to be called with an `ExecuteSqlRequest` without a transaction selector. That will cause Cloud Spanner to use the [default transaction selector](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.ExecuteSqlRequest), which is a single use read-only transaction with a strong timestamp bound.

jiren: Thanks for information.

jiren (approved):

olavloite:
@hengfengli @jiren I don't have permission to merge anything in this repository, so would one of you merge this PR? Thanks!

hengfengli:
@jiren I found the reason why the test fails. The error is: 

```
  1) Google Cloud Spanner API samples write data with array types and read
     Failure/Error:
       write_read_bool_array project_id: @project_id,
                             instance_id: @instance.instance_id,
                             database_id: database.database_id

     ArgumentError:
       wrong number of arguments (given 1, expected 2)
     # ./spanner_samples.rb:1506:in `write_read_bool_array'
     # ./spec/spanner_samples_spec.rb:1456:in `block (3 levels) in <top (required)>'
     # ./spec/spanner_samples_spec.rb:128:in `capture'
     # ./spec/spanner_samples_spec.rb:1455:in `block (2 levels) in <top (required)>'
```

The actual error happens at `client.insert`: 

```rb
  client.insert "Boxes", BoxId: box_id, ErrorChecks: [true, false, true]
```

The above code would not work any more. I believe the reason is the definition of `client.insert` has been updated. We have to change the code to: 

```rb
  client.insert "Boxes", [{BoxId: box_id, ErrorChecks: [true, false, true]}]
```

All calls of `client.insert` need to be updated.

jiren:
> @jiren I found the reason why the test fails. The error is:
> 
> ```
>   1) Google Cloud Spanner API samples write data with array types and read
>      Failure/Error:
>        write_read_bool_array project_id: @project_id,
>                              instance_id: @instance.instance_id,
>                              database_id: database.database_id
> 
>      ArgumentError:
>        wrong number of arguments (given 1, expected 2)
>      # ./spanner_samples.rb:1506:in `write_read_bool_array'
>      # ./spec/spanner_samples_spec.rb:1456:in `block (3 levels) in <top (required)>'
>      # ./spec/spanner_samples_spec.rb:128:in `capture'
>      # ./spec/spanner_samples_spec.rb:1455:in `block (2 levels) in <top (required)>'
> ```
> 
> The actual error happens at `client.insert`:
> 
> ```ruby
>   client.insert "Boxes", BoxId: box_id, ErrorChecks: [true, false, true]
> ```
> 
> The above code would not work any more. I believe the reason is the definition of `client.insert` has been updated (adding `commit_options`?). We have to change the code to:
> 
> ```ruby
>   client.insert "Boxes", [{BoxId: box_id, ErrorChecks: [true, false, true]}]
> ```
> 
> All calls of `client.insert` need to be updated.

@hengfengli  Yes you are right. Following code also work without array
```
 client.insert "Boxes", {BoxId: box_id, ErrorChecks: [true, false, true]}
```

Let me know if you are going to make changes. Or should I create a PR with all changes.

jiren:
> @hengfengli @jiren I don't have permission to merge anything in this repository, so would one of you merge this PR? Thanks!

Merge option only enabled once the test cases passed.  This will be merged once fix for test cases merged #805

hengfengli:
> > @jiren I found the reason why the test fails. The error is:
> > ```
> >   1) Google Cloud Spanner API samples write data with array types and read
> >      Failure/Error:
> >        write_read_bool_array project_id: @project_id,
> >                              instance_id: @instance.instance_id,
> >                              database_id: database.database_id
> > 
> >      ArgumentError:
> >        wrong number of arguments (given 1, expected 2)
> >      # ./spanner_samples.rb:1506:in `write_read_bool_array'
> >      # ./spec/spanner_samples_spec.rb:1456:in `block (3 levels) in <top (required)>'
> >      # ./spec/spanner_samples_spec.rb:128:in `capture'
> >      # ./spec/spanner_samples_spec.rb:1455:in `block (2 levels) in <top (required)>'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The actual error happens at `client.insert`:
> > ```ruby
> >   client.insert "Boxes", BoxId: box_id, ErrorChecks: [true, false, true]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The above code would not work any more. I believe the reason is the definition of `client.insert` has been updated (adding `commit_options`?). We have to change the code to:
> > ```ruby
> >   client.insert "Boxes", [{BoxId: box_id, ErrorChecks: [true, false, true]}]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > All calls of `client.insert` need to be updated.
> 
> @hengfengli Yes you are right. Following code also work without array
> 
> ```
>  client.insert "Boxes", {BoxId: box_id, ErrorChecks: [true, false, true]}
> ```
> 
> Let me know if you are going to make changes. Or should I create a PR with all changes.

Thanks for submitting the PR. To be honest, I would still prefer to use `[]` because it is more clear and no confusion.

PR #813, opened 2021-06-09
docs(run): connect cloud storage to cat_album

trambui09:
Connected Google Cloud Storage with the starter app. This will fit into the docs-sample created for the documentation of deploying a Rails app to Cloud Run (fully managed)


run/rails-cat_album/Gemfile.lock
```diff
@@ -95,15 +127,20 @@ GEM
...
     mini_mime (1.0.3)
-    mini_portile2 (2.5.3)
     minitest (5.14.4)
     msgpack (1.4.2)
+    multi_json (1.15.0)
+    multipart-post (2.1.1)
     nio4r (2.5.7)
-    nokogiri (1.11.6)
-      mini_portile2 (~> 2.5.0)
+    nokogiri (1.11.6-x86_64-darwin)
```
dazuma, line 141: We need to change this not to be darwin-specific. Can you `bundle update --add-platform ruby` and then `bundle update --remove-platform x86_64-darwin`?

run/rails-cat_album/config/environments/development.rb
```diff
@@ -33,7 +33,7 @@
   end
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
-  config.active_storage.service = :local
+  config.active_storage.service = :google
```
dazuma, line 36: It would be better if we could set this to `:local` for development and `:google` for production, so you don't have to hit the google service when testing locally in the development environment. (That would require getting a suitable helper like `url_for` working properly in the templates above. So if it's not immediatly obvious how to get that to work, we can defer it to a follow-up PR. But I do think we need to get it working before we publish.)

dazuma (changes requested):

run/rails-cat_album/config/environments/development.rb
trambui09: good call! For the url_for helper, is that referring to the local development testing?

trambui09:
Made the proper edits requested above

run/rails-cat_album/config/environments/development.rb
dazuma: Yes: in the view, if storage uses `:local` for development, it stores image data on the local disk rather than in cloud storage, so the img tag needs to be different. I'm assuming that `url_for` knows how to generate the correct URL based on the ActiveStorage service (and if not, then the ActiveStorage documentation/guide should tell us how to generate proper views that will work both for development and production.)

run/rails-cat_album/app/views/photos/index.html.erb
```diff
@@ -5,7 +5,7 @@
     <% @photos.each do |photo| %>
       <div class="card" style="width: 18rem;">
         <div class="card-body">
-          <img class="card-img-top" src="<%= url_for(photo.image) %>" alt="Card image cap">
+          <img class="card-img-top" src='<%= "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}" %>' alt="Card image cap">
```
dazuma, line 8: I suspect these need to be updated so they work properly for both development (`:local`) and production (`:google`). It may be that the original `url_for(photo.image)` code is what we want. But please make sure it works for both cases.

trambui09:
Pulled out the views logic into its own helper method to render the image url based on the active_storage configuration

run/rails-cat_album/app/helpers/photos_helper.rb
```diff
@@ -1,2 +1,11 @@
 module PhotosHelper
+  def image_url photo
+    case Rails.application.config.active_storage.service
+    when :local
+      url = url_for photo.image
+    when :google
+      url = "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}"
+    end
+    url
```
dazuma, line 9: Some Ruby tips:

Every Ruby construct has a "return value". In this case, you have a `case` statement, which is the final value of whatever "when" clause gets selected. In other words, what you have above:

```ruby
case Rails.application.config.active_storage.service
when :local
  url = url_for photo.image
when :google
  url = "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}"
end
```

is equivalent to:

```ruby
url = case Rails.application.config.active_storage.service
      when :local
        url_for photo.image
      when :google
        "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}"
      end
```

If you make that change, your method as a whole would look like this:

```ruby
def image_url photo
  url = case Rails.application.config.active_storage.service
        when :local
          url_for photo.image
        when :google
          "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}"
        end
  url
end
```

At that point, you don't need the `url` variable at all anymore because the return value of the method is the return value of the last statement in the method, which is the `case` statement. So this is now equivalent to:

```ruby
def image_url photo
  case Rails.application.config.active_storage.service
  when :local
    url_for photo.image
  when :google
    "https://storage.googleapis.com/cat_album_storage/#{photo.image.key}"
  end
end
```

run/rails-cat_album/config/storage.yml
```diff
@@ -15,11 +15,10 @@ local:
...
-#   service: GCS
-#   project: your_project
-#   credentials: <%= Rails.root.join("path/to/gcs.keyfile") %>
-#   bucket: your_own_bucket
+google:
+  service: GCS
+  project: trambui-test-1
+  bucket: cat_album_storage
 
 # Use rails credentials:edit to set the Azure Storage secret (as azure_storage:storage_access_key)
```
dazuma, line 23: I'd remove all the other non-Google commented examples.

dazuma (changes requested):

run/rails-cat_album/app/helpers/photos_helper.rb
trambui09: Ahh I see, this makes a lot of sense and looks more clean. TY for the tip Daniel!

run/rails-cat_album/app/views/photos/index.html.erb
```diff
@@ -5,7 +5,7 @@
     <% @photos.each do |photo| %>
       <div class="card" style="width: 18rem;">
         <div class="card-body">
-          <img class="card-img-top" src="<%= url_for(photo.image) %>" alt="Card image cap">
+          <img class="card-img-top" src="<%= image_url photo %>" alt="Card image cap">
```
quartzmo, line 8: Have you tried using [image_tag](https://api.rubyonrails.org/classes/ActionView/Helpers/AssetTagHelper.html#method-i-image_tag) here instead of the hardcoded `img` HTML tag to see if it will use the correct ActiveStorage `service` automatically? (If this works you should be able to remove `image_url`.)

trambui09: @quartzmo I've tried using image tag in this capacity `"alt="Card image cap">`, I'm testing out image_tag with `<%= image_tag photo.image, size: "200x200" %>` and it seems to work in local development. I'm going to start the build and deploy to see if it works in production!

trambui09:
Tried out image_tag last week, but it didn't work in production. The recent pushed included the fix for the image_url helper method and cleaned up the storage.yml file. Please let me know if I need to fix anything else before the merge to main.

quartzmo:
> Tried out image_tag last week, but it didn't work in production.

What was the error or behavior? Do you think it could possibly be a bug in Rails?

trambui09:
> > Tried out image_tag last week, but it didn't work in production.
> 
> What was the error or behavior? Do you think it could possibly be a bug in Rails?

In the production website, when I tried to click on the photo URL, it was broken because it pointed to the active storage blob instead of the cloud storage URL. It could be a bug in Rails, but I feel like it's such a common use case for it not to be fixed yet...

run/rails-cat_album/Gemfile
```diff
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby "~> 2.7.2"
-
+gem "dotenv-rails"
```
dazuma, line 5: Specify a version constraint like `"~> 2.7"`

run/rails-cat_album/app/views/photos/index.html.erb
quartzmo: > In the production website, when I tried to click on the photo URL, it was broken because it pointed to the active storage blob instead of the cloud storage URL.

If you want to download the image by clicking on it, I think maybe you want to add a Rails `link_to` helper around the Rails `image_tag` helper? As discussed here: https://stackoverflow.com/questions/43280001/rails-image-as-a-link

dazuma (approved):
Looks great!

PR #1009, opened 2022-09-13
feat(Spanner): add samples for jsonb

NivedhaSenthil:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist
- [ ] **Tests** pass
- [ ] Please **merge** this PR for me once it is approved.


spanner/spanner_postgresql_jsonb_query_parameter.rb
```diff
@@ -0,0 +1,41 @@
...
+  # instance_id = "Your Spanner instance ID"
+  # database_id = "Your Spanner database ID"
+
+  sql_query = <<~QUERY
+    SELECT venueid, venuedetails 
+    FROM Venues
+    WHERE CAST(venuedetails ->> 'rating' AS INTEGER) > $1
+  QUERY
+
+  results = client.execute sql_query, params: { p1: 5 }
```
bajajneha27, line 29: What does the `param` mean here? 
I think it'll be good if we add some comments here.

spanner/spec/spanner_postgresql_jsonb_add_column.rb
```diff
@@ -0,0 +1,40 @@
...
+require_relative "../spanner_postgresql_jsonb_add_column"
+
+describe "Google Cloud Spanner Postgres examples" do
+  before :each do
+    cleanup_database_resources
+  end
+
+  after :each do
+    cleanup_database_resources
+    cleanup_instance_resources
```
bajajneha27, line 26: Just curious, where do we create instance resources?

bajajneha27 (commented):
Overall LGTM. Have some minor comments/questions

spanner/spanner_postgresql_jsonb_query_parameter.rb
NivedhaSenthil: Param here is the key value pair for parameterised query's parameter.
Sure will add a comment here 👍

spanner/spec/spanner_postgresql_jsonb_add_column.rb
NivedhaSenthil: Instance is not created in the spec.. it is passed on via env variable `GOOGLE_CLOUD_SPANNER_TEST_INSTANCE `

NivedhaSenthil: Just realised we don't have to cleanup instance resource here as we don't create any for this particular test. 
Thanks for pointing out dint realise it earlier 👍

bajajneha27 (approved):

PR #1010, opened 2022-09-15
feat(Spanner): add samples for fine grained access control

NivedhaSenthil:
## Description

Fixes #<ISSUE-NUMBER>

## Checklist
- [ ] **Tests** pass
- [ ] Please **merge** this PR for me once it is approved.


spanner/spanner_read_data_with_database_role.rb
```diff
@@ -0,0 +1,37 @@
...
+  # database_id = "Your Spanner database ID"
+
+  role = "new_parent"
+  spanner = Google::Cloud::Spanner.new project: project_id
+  client = spanner.client instance_id, database_id, database_role: role
+
+  result = client.execute_sql "SELECT * FROM Singers"
+
+  result.rows.each do |row|
+    puts "SingerId: #{row[:SingerId]}, FirstName: #{row[:FirstName]}, LastName: #{row[:LastName]}"
```
SandeepTuniki, line 30: Maybe split this into 3 different lines for better readability?

```rb
puts "SingerId: #{row[:SingerId]}"
puts "FirstName: #{row[:FirstName]}"
puts "LastName: #{row[:LastName]}"
```

spanner/spec/spanner_enable_fine_grained_access_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  after :each do
+    cleanup_database_resources
+    cleanup_instance_resources
+  end
+
+  example "Enable fine grained access" do
+    create_singers_albums_database
+    iam_member = "serviceAccount:spanner-samples-test@helical-zone-771.iam.gserviceaccount.com"
```
SandeepTuniki, line 30: Is it ok to include this service account in the source code? Apart from divulging this info publicly, I think this will make the test brittle if the service account were to be ever changed.

spanner/spec/spanner_enable_fine_grained_access_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    capture do
+      spanner_enable_fine_grained_access project_id: @project_id,
+                                         instance_id: @instance_id,
+                                         database_id: @database_id,
+                                         iam_member: iam_member, 
+                                         database_role: "new_parent",
```
SandeepTuniki, line 36: I see some instances where the `database_role` is hardcoded in the sample, whereas this test is passing it as a parameter. Can we follow a uniform convention for this? I'm leaning towards parameterisation approach.

spanner/spec/spec_helper.rb
```diff
@@ -16,6 +16,7 @@
 require "google/cloud/spanner"
 require "google/cloud/spanner/admin/instance"
 require "google/cloud/spanner/admin/database"
+require_relative "../spanner_samples"
```
SandeepTuniki, line 19: [Nit] This is a bit worrying, since it creates a dependency in the wrong direction :) But I see the helpers are using the samples, so it is necessary. We might, however, end up with circular dependency issues in the future.

spanner/spec/spec_helper.rb
```diff
@@ -127,4 +128,110 @@ def create_test_database database_id, statements: []
     job.wait_until_done!
     database = job.results
   end
+
+    # Creates a temporary database with random ID (will be dropped after test)
+  # (re-uses create_database to create database with Albums/Singers schema)
```
SandeepTuniki, line 133: There's a leading space in the first line. The 1st comment-line doesn't quite align with the 2nd comment-line.

SandeepTuniki (changes requested):

spanner/spec/spanner_enable_fine_grained_access_spec.rb
NivedhaSenthil: I had left it that to be consistent with other languages, so that the samples would look same in the doc.

spanner/spec/spanner_enable_fine_grained_access_spec.rb
NivedhaSenthil: Had the same thought but left it as I saw few places where we already have it that way, I will try creating one in the test 👍

spanner/spec/spec_helper.rb
NivedhaSenthil: ya that was a bit cumbersome, but should be fine as we just require the helper in tests.

SandeepTuniki (approved):
A Rubocop error to be fixed. But otherwise LGTM.

spanner/spec/spanner_enable_fine_grained_access_spec.rb
NivedhaSenthil: Have updated the test to create and cleanup service account now. Please review

PR #1023, opened 2022-10-19
sample(spanner): Add samples for dml returning

diptanshumittal:


NivedhaSenthil:
Please create separate files for each sample.

spanner/spec/spanner_dml_returning_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+    cleanup_database_resources
+    create_singers_albums_database
+    insert_data project_id: @project_id,
+                instance_id: @instance_id,
+                database_id: @database_id
+  end
+
+  after :each do
+    cleanup_database_resources
+    cleanup_instance_resources
```
NivedhaSenthil, line 31: Don't think we need to cleanup instances as we dont create any for these tests.

spanner/spec/spanner_postgresql_dml_returning_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+describe "Google Cloud Spanner Postgres DML examples" do
+  before :each do
+    cleanup_database_resources
+    create_spangres_singers_albums_database
+    create_spangres_singers_table
+    add_data_to_spangres_singers_table
+  end
+
+  after :each do
+    cleanup_database_resources
```
NivedhaSenthil, line 30: Same as above we could probably remove `cleanup_instance_resources`

NivedhaSenthil (approved):

PR #1171, opened 2023-09-16
fix: refactor error reporting code sample

minherz:
align code sample for error reporting with code in other languages (b/275730159):
- use the "fixed" error message


dazuma (approved):
Reviewed the internal issue; approving the update to bring the message in alignment with corresponding samples for other languages.

PR #1254, opened 2024-01-10
chore(main): fix for tests failing because of bundler change

shivgautam:


shivgautam:
This change installed the correct version of the bundler (2.2.2) for Ruby 2.7 but we are still seeing the same errors as earlier.

PR #1510, opened 2024-11-20
Update Gemfile with region tags

apoorvasridaran:
## Description

We need this sample to display without the comments in docs.


iennae (approved):

iennae:
As this is just a region tag addition, bypassing branch protections for the test.

PR #1543, opened 2025-02-16
fix: remove manual installation of gcloud cli 

glasnt:
## Description

Changes in https://github.com/googleapis/testing-infra-docker/pull/446 causing all tests to fail on gcloud signup

First commit: replicates issue

Second commit onwards: attempting to resolve issue

Presumes that this is the correct way to mitigate the upstream changes. Build failures match nightly build issues and are unrelated to this change.


dazuma (approved):
Thanks @glasnt. Sorry, didn't realize ruby-docs-samples was installing its own copy.

PR #1580, opened 2025-07-25
fix(rails): update Dockerfile 

kumaripuja-ssk:
- Updated URLs in Docker file as per the requirement.
- Related Buganizer: http://b/433251254

### DIFF
--- a/run/rails/Dockerfile
+++ b/run/rails/Dockerfile
@@ -7,13 +7,28 @@
 FROM ruby:3.2-buster
 # [END cloudrun_rails_base_image]
 
-RUN (curl -sS https://deb.nodesource.com/gpgkey/nodesource.gpg.key | gpg --dearmor | apt-key add -) && \
-    echo "deb https://deb.nodesource.com/node_20.x buster main"      > /etc/apt/sources.list.d/nodesource.list && \
-    apt-get update && apt-get install -y nodejs lsb-release
+# Replace expired Buster repositories with archive.debian.org
+RUN sed -i 's|http://deb.debian.org/debian|http://archive.debian.org/debian|g' /etc/apt/sources.list && \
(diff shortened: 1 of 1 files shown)


run/rails/Dockerfile
```diff
@@ -7,23 +7,38 @@
...
 # Application dependencies
 COPY Gemfile Gemfile.lock ./
 
 RUN gem install bundler && \
-    bundle config set --local deployment 'true' && \
-    bundle config set --local without 'development test' && \
-    bundle install
+bundle config set --local deployment 'true' && \
+bundle config set --local without 'development test' && \
+bundle install
```
glasnt, line 41: nit: It's useful when using command continuations to use indenting to make it clear that this is all one layer, one run command. 

Please apply to other commands.

run/rails/Dockerfile
```diff
@@ -7,23 +7,38 @@
...
+echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/10no-check-valid-until && \
+apt-get update
+
+# Install Node.js (from NodeSource)
+RUN curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | \
+gpg --dearmor -o /usr/share/keyrings/nodesource.gpg && \
+echo "deb [signed-by=/usr/share/keyrings/nodesource.gpg] https://deb.nodesource.com/node_20.x buster main" > \
+/etc/apt/sources.list.d/nodesource.list && \
+apt-get update && \
+apt-get install -y nodejs lsb-release
```
glasnt, line 22: nit: it would make more sense to install the lsb-release package in the same step that the repository is setup

run/rails/Dockerfile
```diff
@@ -7,23 +7,38 @@
 FROM ruby:3.2-buster
```
glasnt, line 7: question: if there's an issue with buster, why not update the base image to bookworm? Does that version have the same issue?

PR #1590, opened 2025-07-31
chore: Update appengine samples to google-cloud-pubsub/v3.x

aandreassa:
Follows https://github.com/googleapis/google-cloud-ruby/blob/main/google-cloud-pubsub/OVERVIEW.md


dazuma (approved):
Thanks!

PR #1777, opened 2026-04-09
chore(spanner): samples to set transaction options

skuruppu:
Adds samples that show how to set the isolation level and read lock mode at the Spanner client level and at a transaction level. If it's set at the client level, all RW transactions will set this value. The transaction-level setting allows users to override the client-level setting or to set the values for specific transactions.


skuruppu:
It looks like the Ruby client may not support setting the isolation level and lock mode at a client-level. I will have to send a separate PR to ruby-spanner to add this support.

aandreassa:
Let's retry now that https://rubygems.org/gems/google-cloud-spanner/versions/2.36.0 is out!

@skuruppu

skuruppu:
> Let's retry now that https://rubygems.org/gems/google-cloud-spanner/versions/2.36.0 is out!
> 
> @skuruppu

Thanks @aandreassa, the tests now pass.

aandreassa (approved):

